PR #27, opened 2026-05-23
feat: compose self-host AGENTS with Codex

eugenelim:
## Summary
- aggregate Codex skill descriptions across packs before splicing AGENTS.md
- compose self-host AGENTS.md from the core body seed, generated Codex block, and footer
- close self-hosting AC8 in the spec/roadmap while leaving comparison-rule strengthening open

## Validation
- PYTHONPATH=packages/agentbundle python3 -m unittest packages.agentbundle.agentbundle.build.tests.test_adapter_codex packages.agentbundle.agentbundle.build.tests.test_self_host_check
- make build-check
- tools/lint-build.sh
- tools/test-all.sh
- git diff --check


eugenelim (commented):
Nice unlock — multi-pack Codex aggregation lets `make build-self` finally project an honest AGENTS.md, and the spec/plan/roadmap updates land cleanly with the code. A few things to weigh before merging; the first is the biggest.

---

### 1. Subagent guidance silently disappears from AGENTS.md  *(major)*

`packs/core/seeds/AGENTS.md:119–122` & deletions in `AGENTS.md:107–143`

The old AGENTS.md had ~30 lines under **Specialist subagents** naming `adversarial-reviewer`, `security-reviewer`, `quality-engineer`, and `implementer` with the lens each one applies. That guidance is now gone from the projected AGENTS.md entirely. The generated `<!-- agent-skills:start -->` block aggregates **skills only** (it walks `.apm/skills/`), and the replacement section is a one-liner that just says "reviewers live under `.claude/agents/`".

Net effect: an agent reading AGENTS.md no longer learns *when* to invoke which reviewer — arguably the most useful 30 lines of the old file. Options:

- Keep a curated subagent bullet list in the seed body (parallel structure to the generated skill block, just authored not generated).
- Or extend the codex adapter to also aggregate `.claude/agents/*.md` frontmatter into a second managed block.

Removing it silently as part of a composition refactor reads as scope creep relative to what the PR title advertises ("compose self-host AGENTS with Codex").

### 2. "Generated skill list **below**" is nowhere near the section that says so  *(medium)*

`packs/core/seeds/AGENTS.md:121` / composed `AGENTS.md:115–151`

The new section says *"Use the generated skill list below…"* — but the list actually lands at the very end of the file, after `## Check before acting`, `### Excuses we don't accept`, `## When this file is wrong`, and the horizontal-rule + `*Generated from…*` line. A reader skimming the section sees an empty heading and has to scroll past the entire rest of the document to find what "below" refers to.

Small fix: drop `<!-- agent-skills:start -->` / `<!-- agent-skills:end -->` markers in the seed *immediately under this paragraph*. `_project_managed_block` already splices between existing markers when it finds them and only appends when they're missing — so once the markers exist in the seed, the list lands where the prose promises and the body's natural footer stays at the bottom.

### 3. `codex.project_packs` runs twice per self-host build  *(medium)*

`packages/agentbundle/agentbundle/build/self_host.py:180–181` and `:204` (inside `_compose_agents_md`)

`_project_all_adapters` now calls `codex.project_packs(...)` once across all packs. Then `_compose_agents_md` immediately overwrites AGENTS.md with the body seed and calls `codex.project_packs(...)` **again**. The first call's output into the working-tree AGENTS.md is unconditionally clobbered by the body-seed write that opens composition — it's dead work.

Two clean options:
- Drop the `if adapter_name == "codex"` branch in `_project_all_adapters` and let `_compose_agents_md` own the codex projection (composition is now the only legitimate consumer of multi-pack codex under self-host).
- Or keep it there and stop re-running codex inside `_compose_agents_md`.

Either way, one of the two calls is redundant and the duplication will rot.

### 4. Stale Phase-2 comment in `self_host.py`  *(low)*

`packages/agentbundle/agentbundle/build/self_host.py:218–223`

The header block above `EXCLUDED_PATTERNS` still says *"AGENTS.md body+footer composition remain Phase 2"*. That's exactly what this PR lands. Spec/plan/roadmap got updated for the close; this comment didn't. (The `Comparison-rule strengthening` half is still accurate.)

### 5. `AGENTS.local.md` is a pointer to nothing  *(low)*

`AGENTS.local.md:1–5`

The whole file is self-description: *"Repo-local addendum… keep guidance here specific to this repository."* The composed AGENTS.md footer now points readers here (`> Working on this repo specifically? See AGENTS.local.md.`), so anyone following the pointer lands on a placeholder with no actual local guidance.

Either (a) seed it with at least one real repo-local note adopters can see and replace, or (b) make the footer pointer conditional on the file having content beyond boilerplate. Shipping the pointer + the empty target together teaches readers to ignore the pointer.

### 6. Bootstrap-target test loses subagent assertions  *(low, but coupled to #1)*

`tools/test-bootstrap-targets.sh:98–119`

The deleted blocks were the only assertions pinning the `adversarial-reviewer`, `security-reviewer`, and `quality-engineer` bullets in AGENTS.md. Combined with the seed no longer carrying that content, nothing — neither a unit test nor a bootstrap check — now pins that those subagents are documented to readers of AGENTS.md.

If the decision is "subagents stay discoverable only via `.claude/agents/`", that's a real change to how the surface is documented and should be called out in the spec / CHARTER rather than folded into a composition PR. If they should still be in AGENTS.md, restore the assertions pointed at wherever the content ends up.

eugenelim:
2. Fixed in eaa60ab. Moved the Codex managed-block markers directly under the Agent workflows paragraph, so the generated skill list now appears where the section points readers.

eugenelim:
1. Fixed in eaa60ab. Restored the curated Specialist subagents section in the AGENTS seed and regenerated root AGENTS.md so the reviewer guidance is visible again.

eugenelim:
3. Fixed in eaa60ab. Removed the redundant Codex projection from _project_all_adapters; _compose_agents_md now owns the single aggregate Codex splice after writing the body seed.

eugenelim:
5. Fixed in eaa60ab. Added concrete repo-local guidance to AGENTS.local.md: maintainers should edit pack-side sources first, run make build-self, then verify with make build-check.

eugenelim:
4. Fixed in eaa60ab. Updated the stale self_host.py header: AGENTS.md composition is no longer described as remaining Phase 2 work; only comparison-rule strengthening remains open.

eugenelim:
6. Fixed in eaa60ab. Restored bootstrap-target assertions for the adversarial, security, and quality reviewer bullets, now pointed at the restored Specialist subagents section.

PR #621, opened 2026-07-23
feat(gates): wire validate-claude-plugin-manifests into local build-check chain

eugenelim:
## Summary

- Adds `tools/validate-claude-plugin-manifests.py` as the 10th and final step in `build_gate_chain.py`'s `build_check` function
- Malformed `dist/claude-plugins/*.claude-plugin/plugin.json` manifests and `marketplace.json` are now caught locally (`make build-check`) rather than only at publish-time CI
- Updates unit test: step count 9 → 10, expected spawned-path list extended, "nine" comment updated to "ten"
- Updates module docstring, `build_check` docstring, and subparser help to enumerate the new step

**Why:** `validate-claude-plugin-manifests.py` already ran as a CI gate in `publish-claude-plugins.yml` but was not wired into the local `make build-check` chain, creating a gap where a manifest error could only be caught at publish time. The `dist/claude-plugins/` dependency is satisfied by the `cmd_build` step that runs earlier in the same chain.

**Notes:** The validator hardcodes `dist/claude-plugins/`; documented in spec Assumptions as a default-output-dir-only coupling.

## Test plan

- [x] `python tools/test_build_gate_chain.py` exits 0 (8 tests pass)
- [x] `python tools/validate-claude-plugin-manifests.py` exits 0 (all 20 manifests pass)
- [x] `make build-check` exits 0 (validator runs successfully as step 10)
- [x] Adversarial reviewer: Concerns addressed (test comment, docstrings, output-dir coupling documented)

### DIFF
--- a/docs/specs/local-claude-manifest-validator-gate/plan.md
+++ b/docs/specs/local-claude-manifest-validator-gate/plan.md
@@ -1,0 +1,89 @@
+# Plan: local-claude-manifest-validator-gate
+
+- **Spec:** [`spec.md`](spec.md)
+- **Status:** Draft
+
+## Constraints
+
+- `tools/build_gate_chain.py` `build_check` uses `_script_step(label, *path_parts)` — consistent with all existing script steps.
+- The test file `tools/test_build_gate_chain.py` asserts exact step count and exact ordered list of spawned paths — both must be updated atomically.
+
+## Risks
+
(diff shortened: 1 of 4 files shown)


eugenelim:
Closing in favour of #622, which covers this work plus . The validator step is placed immediately after `build` (not last) so a manifest failure stops the chain before the slower downstream gates run — semantically closer to what the spec requires. The test assertions in #622 reflect that ordering.

PR #742, opened 2026-07-27
fix(tests): skip credbroker-dependent tests when credbroker is absent

eugenelim:
## Summary

- **Unit tests** (`test_ac18_argv_ban_refused`, `test_ac18_no_token_in_stdout`, `test_ac19_reserved_sso_namespace_refused`): replaced `pytest.importorskip("credbroker")` with a subprocess-based check (`subprocess.run([sys.executable, "-c", "import credbroker"], ...)`). `importorskip` was insufficient because `test_linear_primitive.py` injects a credbroker stub into `sys.modules` earlier in the full suite via `sys.modules.setdefault`, causing `importorskip` to find the stub and not skip — while the subprocess still lacked the real package. The subprocess check spawns a fresh interpreter, bypassing `sys.modules` state.
- **Integration test** (`test_entry_point_imports_resolve_under_user_scope_layout`): extended `_assert_no_relative_import_error`'s `custom_dep_guard` to recognise `"credbroker not found." + "pip install"` as an out-of-scope precondition failure, parallel to the existing `"error: missing dependency"` guard. Only `setup.py` emits this string.
- Removed `workspace.toml [backlog].open` entry `pre-existing-credbroker-test-not-installed`.

## Test plan

- [x] `test_linear_primitive.py` + `test_credential_setup_skill.py` in order (simulates full-suite `sys.modules` pollution): 6 passed, 3 skipped — no failures
- [x] All 20 tests across both affected files: 11 passed, 9 skipped — no failures
- [x] Full suite (`python -m pytest packages/agentbundle/ -q`) running in background

### DIFF
--- a/packages/agentbundle/tests/integration/test_credential_user_scope_invocation.py
+++ b/packages/agentbundle/tests/integration/test_credential_user_scope_invocation.py
@@ -130,9 +130,14 @@ def _assert_no_relative_import_error(result: subprocess.CompletedProcess, entry:
         and not any(mod in stderr for mod in credential_area_modules)
     )
     # Scripts with custom dependency guards emit "error: missing dependency '<pkg>'"
-    # instead of ModuleNotFoundError — same out-of-scope category.
+    # instead of ModuleNotFoundError — same out-of-scope category. setup.py emits
+    # "credbroker not found. …" when credbroker is absent (precondition failure,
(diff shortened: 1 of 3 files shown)


eugenelim:
Superseded by #746 which includes these credbroker skip guards plus the Windows portability fixes.

PR #845, opened 2026-08-03
docs(agents): make packages/AGENTS.local.md a conditional load instruction

eugenelim:
## Summary

- Rewrites the `packages/AGENTS.local.md` pointer in root `AGENTS.local.md` from an informational navigation hint into an imperative conditional load instruction
- Agents running from the repo root don't auto-load subdirectory AGENTS files; the pointer was silently ignored, leaving the two-step closeout rule invisible during work-loop runs that touch `packages/`

## Deferred

The two-step closeout rule itself stays in `packages/AGENTS.local.md` — no content moved.


eugenelim:
Wrong branch — reopening from a clean branch

PR #884, opened 2026-08-07
docs(install): use `python -m pip` for every install command

eugenelim:
## What changed and why

Every runnable install command **for this project** — editable local installs
and the front-door PyPI one-liners — moves from `pip install …` to
`python -m pip install …`.

Bare `pip` resolves through whatever shim is first on `PATH`, which need not
belong to the interpreter that will later run `agentbundle` — the classic
"installed it, but the command isn't there" failure. `python -m pip` binds the
installer to the interpreter by construction.

`python`, not `python3`: Windows ships `python` and generally has no `python3`,
and this repo runs Windows CI (`build-check-windows.yml`). `python3` would be a
cross-OS regression, not a fix. The `guides/figma/` tutorials already used this
form, so the idiom has precedent here.

Two commits:

1. **Editable installs** — 20 files: `AGENTS.md`, the `Makefile` requirement
   comment, `docs/architecture/catalogue.md`, the hand-authored docs-site
   install page, five guides, two credbroker READMEs, and 16 `run:` steps
   across eight CI workflows.
2. **Front-door one-liners** — 18 files: `README.md`, `CONTRIBUTING.md`, the
   marketing site's `InstallTerminal.astro`, three docs-site pages, two
   maintainer guides, nine adopter guides, and `packages/credbroker/README-pypi.md`.

## How it was verified

- `make build-check` — **complete, SAST/SCA included** — exit 0 on both commits.
  `lint-ci-parity` passes: 50 steps dispositioned, 73 extracted targets
  corroborated.
- `make lint-ruff` exit 0. `make build-self FORCE=1` produces no projection drift.
- `test_scaffold_projection.py` passes — the scaffold copies stay byte-identical.
- Line-by-line comparison against `HEAD` confirms all eight
  descriptive-prose lines are byte-identical.
- `pip install --dry-run` comparison confirms `--upgrade` produces an identical
  resolution plan for a path requirement, so it was left out.

**Gate-ordering note:** `lint-catalogue-curation-guard` reads
`git diff origin/main...HEAD` — *committed* state. Run against an uncommitted
working tree it gates an empty changeset and passes vacuously. Any local
`build-check` meant to predict CI has to run after the commit. That is how the
first commit's `packages/agentbundle/README.md` violation was caught.

## What is deliberately unchanged

- **`packages/agentbundle/README-pypi.md`** (and the whole
  `packages/agentbundle/**` tree). RFC-0059 D6's curation guard requires an
  `Engine-Change-RFC:` trailer there. This sweep has no RFC behind it, and
  asserting one to clear a gate would be a false claim. Carving a docs
  exception into the guard was considered and rejected — that is a change to an
  RFC-governed gate and does not belong in a doc PR.
- **`guides/_shared/reference/catalogue-ci-contract.md`.**
  `sync_authoring_scaffold.py` holds it byte-identical to a copy inside
  `packages/agentbundle/_data/catalogue-scaffold/`, enforced by
  `test_scaffold_projection.py` — so editing the guide writes into the guarded
  tree. Same for `catalogue-authoring-standards.md`.
- **The third-party dependency family** — `pip install -r tools/requirements.txt`,
  `pip install ruff mypy`, `pip install docling`, `pip install 'httpx>=0.27'`.
  Those install *prerequisites*, not this project; different decision, different
  blast radius. A mid-loop attempt to sweep just the `-r` form in `AGENTS.md`
  and `CONTRIBUTING.md` was reverted because it left five other sites of the
  same command untouched — the exact cross-file drift this sweep removes.
- **Prose that names the mechanism** rather than giving a command — e.g. "an
  editable clone (`pip install -e`) defaults to itself".
  `docs/architecture/catalogue.md:66` quotes `source_defaults.py`'s error string
  and must stay byte-identical to it.
- **`docs/adr/`, `docs/rfc/`, `docs/specs/`** — frozen records of what was
  decided and run at the time. Rewriting them would make the record lie.
- **Every shipped artifact.** No file under `packs/**` or
  `packages/*/{agentbundle,credbroker}/`, no version bump, no changelog entry.
  A doc update must not cut a release.

## Deferred

- **The two runtime strings that print an install command** —
  `initialise_self_hosted.py` and credential-setup's `setup.py`. Both were
  implemented in an earlier pass and reverted: the first wants a changelog entry
  and a release; the second forces a `credential-brokers` version bump across
  `pack.toml`, `plugin.json`, and `marketplace.json`, and that bump turns
  `test_credential_brokers_pack_install.py` red — concrete evidence the change
  is release-shaped, not doc-shaped. Handed off in
  `docs/specs/pip-invocation-idiom/handoff-runtime-hint-strings.md`.
- **`packages/agentbundle/README-pypi.md`** — needs either an RFC or a
  deliberate change to the curation guard. Same handoff.

  Until these land, the shipped hint strings and the PyPI readme disagree with
  the guides. That is named honestly rather than hidden.

## Review note

`adversarial-reviewer` was a **named skip** — this session's policy prohibits
subagent dispatch unless explicitly requested. The spec-less self-review
checklist was run inline instead; it caught one real defect in the first pass
(a negative test assertion that silently weakened when the matched string grew
a prefix).


eugenelim:
Force-pushed one correction.

`lint-catalogue-curation-guard` failed on CI while passing locally. Two things:

1. **The guard only sees committed changes.** Its path-gate reads `git diff --name-only origin/main...HEAD`, so running `make build-check` against an uncommitted working tree gates an empty changeset and passes vacuously. Local build-check has to run *after* the commit to predict CI.
2. **`packages/agentbundle/README.md` was the violation.** RFC-0059 D6 protects the whole `packages/agentbundle/**` tree behind an `Engine-Change-RFC:` trailer (carve-outs only for `build/recipes/**` and `.../tests/...`).

Dropped that one file from the changeset rather than adding a trailer asserting an RFC this change does not have. `packages/agentbundle/README.md` therefore keeps bare `pip install -e` at lines 8 and 219 while every other doc moves — a known, recorded inconsistency, not an oversight.

**Second open decision for the reviewer:** should the guard grow a docs carve-out (`packages/agentbundle/README*.md`, `**/*.md`)? A README is not engine behaviour, which is what D6 is protecting. That is a change to an RFC-governed guard, so it belongs in its own PR, not here.

PR #894, opened 2026-08-07
docs(rfc): RFC-0082 — engine export boundary and per-surface test inclusion

eugenelim:
## What

Opens **RFC-0082** (Draft) defining the agentbundle engine's runtime export
boundary and its per-artifact test-inclusion rule — the engine-side companion to
ADR-0071, which settled the same question for packs and scoped itself there.

Docs only. No code, no packaging config, no packs, no version bump.

## Why

ADR-0071 made `.apm/` the runtime export boundary for packs. Nothing ever
defined what the *wheel* exports. The result is four distribution surfaces
behaving four different ways — measured against 0.29.8 by building both
artifacts from a clean source copy:

| Surface | Test content today |
| --- | --- |
| zipapp | none — correct, via an undocumented one-liner |
| wheel | 45 of 184 entries |
| sdist | the in-package tree's `.py` modules only (fixtures absent), plus 8 top-level files and no `conftest.py` — present but unrunnable |
| vendored | the whole working tree, unfiltered, plus a pack's tests |

A disposition record from the day before names the gap outright:
*"`agentbundle/build/tests/` is inside the package and does ship — nothing moved
out of there."*

## What it proposes

- **D1** — `packages/<pkg>/<pkg>/` is the boundary; relocate the test tree.
  **The importable package does not move**; only `build/tests/` does.
- **D2** — per-surface inclusion: sdist yes and complete, wheel/zipapp/vendored
  no. The asymmetry is argued from PyPA guidance and redistributor practice, with
  the dissent engaged rather than omitted.
- **D3** — the vendored copy is wheel-class because its documented consumption
  path is `pip install -e`; adopter assurance comes from `catalogue verify`.
- **D4** — `tools/test*.py` co-location recorded as a named exception.
- **D5** — two in-repo pure-stdlib enforcement instruments.

## Evidence

Both artifacts were built and inspected; the relocation was applied to a
temporary copy, installed, and run against an unmoved control. `pip install -e`
is unaffected. Nothing was written into the working tree during the spike.

`docs/rfc/0082-notes/` carries the enforcement-tool transcripts, because D5 is a
designated review focus and its conclusion contradicts the received wisdom:
`check-wheel-contents` does not fire on a nested test tree, and `pydistcheck`'s
natural-reading `--expected-directories` flag **passes while the property is
violated** (setuptools wheels carry no directory entries).

## Review notes

Four adversarial rounds ran against this draft. Findings that changed the
substance, not just the prose:

- The wheel defect is **PEP 420 namespace discovery** under the default
  `namespaces = true` — not the tree's `__init__.py`. Deleting that file leaves
  44 of 45 entries. An earlier draft had this wrong.
- The move would **silently disable a shipped destructive-write guard**
  (`self_host.py`'s `"tests/fixtures/"` substring test) while its covering test
  stays green. Now in the migration and the pre-mortem.
- `agentbundle init --tooling vendored` also ships **pack** tests via a raw tree
  copy — a real hole in ADR-0071 that its projection-adapter reasoning doesn't
  cover.
- Two of the supporting survey's claims did not survive citation re-fetching and
  are corrected in place; the Fedora strand is downgraded to corroborating,
  because its canonical page could not be retrieved to confirm the wording.

RFC-0081 is reserved for an in-flight draft authored elsewhere; the gap in the
sequence is intentional and noted in the index.

## Next

Awaiting an accept/reject decision. On acceptance, the companion ADR and the
follow-on implementation spec follow in the same session — no ADR is authored
against an unaccepted RFC.

### DIFF
--- a/docs/rfc/0082-engine-export-boundary-and-test-inclusion.md
+++ b/docs/rfc/0082-engine-export-boundary-and-test-inclusion.md
@@ -1,0 +1,899 @@
+# RFC-0082: Engine export boundary and per-surface test inclusion
+
+- **Status:** Draft
+- **Author:** eugenelim
+- **Approver:** eugenelim
+- **Date opened:** 2026-08-07
+- **Date closed:**
+- **Decision weight:** heavy
+- **Related:** [ADR-0071](../adr/0071-pack-runtime-export-boundary-and-test-placement.md)
+  (the pack-side companion this extends to the engine),
+  [`docs/specs/pack-test-boundary-remaining-packs/`](../specs/pack-test-boundary-remaining-packs/)
+  (whose disposition record names the gap this closes),
+  [`0082-notes/`](0082-notes/) (promoted spike transcripts)
+
+## Reviewer brief
+
+- **Decision:** whether `packages/<pkg>/<pkg>/` is the engine's runtime export
+  boundary, with test inclusion decided per distribution surface rather than
+  uniformly.
+- **Recommended outcome:** accept.
+- **Change if accepted:**
+  - `packages/agentbundle/agentbundle/build/tests/` relocates to
+    `packages/agentbundle/tests/build/`. The importable package does not move.
+  - The source distribution carries the complete test tree; the wheel, the
+    zipapp, and the vendored copy carry none.
+  - Two pure-stdlib enforcement instruments, both in-repo: an artifact gate
+    over the built wheel and sdist, and a unit test over the vendored payload.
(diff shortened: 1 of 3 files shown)


eugenelim:
Superseded by #895. Same RFC number, widened scope: review found that the engine-only boundary this PR drew does not decouple catalogue tests from the engine folder — 82 modules inside `packages/agentbundle/` assert things about the catalogue, which also made this PR's sdist rule incoherent. Reopened on `eugene/test-ownership-boundaries-rfc` so the branch name matches the content.

PR #932, opened 2026-08-12
feat(guides): remove repo-only governance references and guard them out

eugenelim:
## What does this change?

Shipped `guides/` no longer links or cites into repository-only governance
records — `docs/adr/`, `docs/rfc/`, `docs/specs/`, and changelogs — which
adopters never receive, so those references dangled in every installed copy. A
new pure-stdlib lint, wired into documentation CI, stops them coming back.

## Why?

- Implements: `docs/specs/governance-guides-cleanup/spec.md` (13 ACs / 6 tasks, full mode, registered under `ini-007`)

Removing the references without a guard just resets the clock — they
reappeared with each downstream migration. The guard is the regression fence;
the scrub is the one-time human pass it cannot replace.

The interesting design choice is rule 3. Rather than allowlisting, it reads the
real spec-slug set from `docs/specs/` at runtime, so a citation of a *real*
record fails while invented tutorial slugs (`docs/specs/webhook-retries/`),
concept tokens (`spec/plan`), and `<feature>` placeholders pass
deterministically. The header documents the hole this leaves: a citation of a
spec that does not exist yet is invisible to rule 3, which is why the scrub
needed judgment rather than just the tool.

## How do I verify it?

```
python3 tools/lint-guides-no-repo-only-refs.py                  # prints OK, exits 0
python3 -m pytest tools/test_lint_guides_no_repo_only_refs.py -q   # 18 passed
python3 tools/validate_guides.py                               # OK (0 errors)
python3 tools/check-guide-index.py                              # OK
python3 -m pytest tools/test_build_site_link_rewrites.py \
  tools/test_validate_guides.py tools/test_check_guide_index.py -q   # 30 passed
make build-check                                                # complete, SAST/SCA included
make lint-ruff && make lint-mypy                                # clean
```

All run green locally. `validate_guides` emits 142 pre-existing migration
warnings on packs this PR does not touch.

- **`packs/user-guide-diataxis/`** — guides now point at
  `product-documentation`, but retiring the back-compat shim from the catalogue
  is an RFC-scoped decision, deliberately out of scope.
- **No escape markers used.** The `<!-- guides-lint: allow -->` hatch exists and
  is tested, but every real reference was fixed rather than suppressed.
- **No pack version bump.** The change is guides plus repo-only tooling, so
  nothing ships a new version; the changelog entry says so explicitly.
- **Only scrub-touched guides gained frontmatter.** The other 142 unmigrated
  guides are left for the migration that owns them.
- **Broadened CI trigger accepted.** The docs job previously keyed only on
  `guides/AGENTS.md`; it now runs on `guides/**`, which intentionally widens
  documentation-job coverage.

## Review round 2

The original loop closed with `Clean — ready to commit.`, but two lenses had
never been applied. Running them found real defects, so that verdict covered
less than it appeared to. Full findings and dispositions:
`docs/specs/governance-guides-cleanup/notes/adversarial-review-2.md`.

The two that mattered most:

- **`tools/lint-plugin-route-docs.py` had been weakened so a scrubbed guide
  would pass** — the spec's own `Never do`. `user-guide-diataxis` was dropped
  from the packs `install-routes.md` must enumerate, which made that
  enumeration false: the pack still declares `allowed-scopes = ["repo"]`.
  Reverted, and the name restored to the guide.
- **The 18 tests pinning AC1–AC6 ran in no gate at all.** Not in the `Makefile`
  `test` target, not in `build-check.yml`; the new `docs.yml` job runs the
  linter, not its tests. A future weakening of any detection rule would have
  shipped green. Registered in both, as AC13.

## Known gaps, deliberately not closed here

- **Rule 3 is allow-by-default in both directions.** A citation of a spec that
  does not exist yet is invisible; and an untouched guide turns red the day
  someone creates a `docs/specs/` directory colliding with an invented tutorial
  slug. Inverting the `spec/` arm to a teaching-slug allowlist closes both, but
  changes AC4's detection contract and deserves its own spec.
- **Rule 1 does not see raw HTML anchors, autolinks, or bare unlinked paths in
  prose; rule 2 is case-sensitive.** All documented in the module header rather
  than assumed covered.
- **A symlinked ancestor of `--guides-root` is followed.** I implemented the
  canonical-path check that closes it and reverted it — it also rejects ordinary
  paths wherever a system directory is itself a link, which is every macOS
  `/var` path. AC1 now states what the code enforces.
- **Escape-hatch suppressions are silent.** No marker exists in the repo yet;
  worth emitting a stderr note before the first one lands.

## Provenance

This branch is a reconstruction. The original worktree was destroyed with the
work uncommitted; every edit here was replayed from the original session's
`apply_patch` stream, which was verified to be the complete set of file
mutations for that session. Three hunks did not replay: one whitespace-only
no-op, one superseded by a later richer edit, one already applied. Two
duplicate applications (a repeated regex and a repeated test function) were
deduplicated — `ruff` caught the second, which `pytest` had silently shadowed.

---

## Checklist

- [x] Tests pass locally (see commands above)
- [x] Lint and typecheck pass (`make lint-ruff`, `make lint-mypy`)
- [x] Self-review run via the relevant reviewer subagent — `adversarial-reviewer`, `quality-engineer`, and `security-reviewer` all run on this branch in review round 2 (`notes/adversarial-review-2.md`). **This closed the two named skips the original loop recorded** (`quality-engineer` runtime failure, experience role unavailable). Round 2 found and fixed a spec `Never do` violation, a broken sentence, a destroyed supersession statement, a test suite wired into no gate, and a fail-open junction check.
- [x] Spec and code agree — the changelog edit required widening the spec's `docs/` boundary, done in this PR with AC12 added
- [x] Living docs match reality:
  - [x] `docs/product/changelog.md` updated
  - [x] `guides/` updated — this PR *is* the guide change
  - [ ] `docs/architecture/` — n/a, no structural change
  - [ ] `docs/product/roadmap.md` — n/a, completes no roadmap item
- [x] Learnings captured — the reference-definition finding was generalized into the guard design and its regression test rather than left as a one-off fix


eugenelim:
CI note: **`Gate E — disconnected smoke` is a pre-existing failure on `main`, not introduced here.**

The same job fails on `main` at `88449dfa` — run [31625311907](https://github.com/eugenelim/agent-ready-repo/actions/runs/31625311907) — at the same step, `Configure local source and list-packs`. That step exercises agentbundle offline catalogue installation; this PR touches `guides/`, two repo-only lint tools, and spec documents, none of which it reads.

Everything else is green, including `make build-check` (locally run to completion **with** the SAST/SCA leg, since the diff touches `tools/`), `build-check (windows)`, `codeql`, `ci-security`, and the new `Shipped guides contain no repo-only governance references` job.

PR #980, opened 2026-08-16
feat(sast): make the suppression-equivalence check a tool

eugenelim:
## What does this change?

Adds `tools/compare-bandit-suppressions.py <base-ref> [<head-ref>]` and its self-test. Run it whenever a `# nosec` is touched.

## Why?

Rewriting a suppression is supposed to be behaviour-preserving, and the only way to know is to compare what Bandit reports before and after. That procedure exists — in `bandit-nosec-comment-hygiene`'s § Testing Strategy — but it lives as **prose inside a Frozen spec**, so when the procedure improves the record cannot be corrected. That is the defect this closes.

It earns being a tool for a second reason: the procedure has four traps, and **each gave a wrong answer the first time it was run by hand**. All four fail *silently* — a confident wrong number, not an error.

| Trap | What went wrong | Where it lives now |
| --- | --- | --- |
| Rows are not keys | 362 rows collapse to 239 distinct keys | `scan()` reports both, diffs only keys |
| Two clean worktrees | worktree vs. *dev tree* scanned gitignored build output on one side, inventing a 27-key delta | both sides `git worktree add --detach` |
| Relative scan roots | absolute paths stop `exclude_dirs` matching, so the scans cover different file sets | roots relative, `cwd` is the worktree |
| Both scans exit 1 | findings are expected at low/low, so a non-zero exit aborted the run | exit code ignored; only a missing report is fatal |

A **fifth** surfaced while building it: `Path.resolve()` anchors a relative path to the calling process's cwd — the repo, not the worktree — silently mixing trees. `_worktree_relative` is string-based for that reason.

Two checks, because neither suffices alone: the reported-finding diff catches a suppression that weakened or widened onto a test that fires somewhere; the resolved-id inventory (through Bandit's own `_parse_nosec_comment`) catches what the scan cannot see — widened onto a test that fires nowhere, or resolving to no id at all.

Spec: `docs/specs/compare-bandit-suppressions/spec.md`

## How do I verify it?

Validated against the change that motivated it, which is the acceptance test — the tool is only worth having if it agrees with the hand-run procedure it replaces:

```
python3 tools/compare-bandit-suppressions.py 1f6b2d2f 7f186a0b
```

```
  base (1f6b2d2f): 362 rows -> 239 distinct keys, 20 suppressions, 54 stderr line(s)
  head (7f186a0b): 362 rows -> 239 distinct keys, 23 suppressions, 0 stderr line(s)

reported findings: identical (239 distinct keys)
  note: new file at head carries 2 suppression(s): tools/run-bandit-gate.py
  note: new file at head carries 2 suppression(s): tools/test-sast-stderr-gate.py

RESOLVED SUPPRESSION IDS DIFFER:
  tools/capture-publish-control-evidence.py:
      base [frozenset({'B310'}), frozenset({'B310'})]
      head [frozenset({'B310'})]
```

Every number matches what the spec records — **362 → 239 both sides, empty symmetric difference, stderr 54 → 0**. The one reported difference is exactly the spurious prose-block directive that spec's AC2 removed. **Detecting an intended change is the tool working**: exit 1 means "look", not "you broke it".

```
python3 tools/test-compare-bandit-suppressions.py         # fast trap cases
python3 tools/test-compare-bandit-suppressions.py --e2e   # + same-ref comparison, exits 0
```

That validation run also **found a flaw in the first version**: it counted suppressions in files present at only one revision as differences, and reported a false FAIL because the change under test added two files. Only a file present at both revisions can have had its suppressions changed; added and removed files are now notes.

- **Wiring it into `make build-check`.** It creates two worktrees and runs two full scans. It is a harness for the human touching a suppression, not a per-PR gate. Its self-test's fast cases join `tools/test-all.py`'s curated manifest instead.
- **Pure stdlib.** AGENTS.md requires it for new `tools/` scripts; this one is a Bandit *driver* — it shells out to `bandit` and imports its parser — so the rule cannot apply. Named in § Declined rather than worked around.
- **Raising the severity floor to match the gate's.** `low/low` is deliberate: a suppression that moved a low-severity finding is still a moved suppression, and `medium/medium` would hide it.

## Note on ordering

`compare-bandit-suppressions-tool` is added to `workspace.toml [backlog].open` by the sibling PR #977 and does not exist on the `origin/main` this branch is cut from, so this PR cannot remove it. **Whichever of the two merges second removes it.** Stated rather than silently checked — an AC claiming a deletion that never happened is worse than an open one.

### DIFF
--- a/docs/specs/compare-bandit-suppressions/spec.md
+++ b/docs/specs/compare-bandit-suppressions/spec.md
@@ -1,0 +1,188 @@
+# Spec: compare-bandit-suppressions
+
+- **Status:** Shipped
+- **Owner:** eugenelim
+- **Plan:** none — single task; verification mode is named in § Testing Strategy
+- **Mode:** full (governance surface — it is the verification procedure for
+  changes to `# nosec` suppressions, which govern what the SAST gate skips)
+- **Constrained by:** [ADR-0084](../../adr/0084-nosec-reason-delimiter-and-stderr-as-a-gate.md),
+  [ADR-0017](../../adr/0017-adopt-bandit-pip-audit-semgrep-sast-gate.md)
+- **Contract:** none (a manual verification harness; no published interface)
+- **Shape:** service
+
+> **Spec contract:** this document defines what "done" means. The implementing
+> PR must match this spec, or update it. Verification must be derivable from it.
+
+## Named deviation from full mode
+
+No `loop-engine` / `loop-cohort` state machine and no `plan.md` — a single task
+whose verification is the tool's own reproduction of a historical result. The
+two human approval gates were **granted up front by the requester**.
+
+## Objective
+
+Rewriting a `# nosec` comment is supposed to be behaviour-preserving, and the
+only way to know is to compare what Bandit reports before and after. That
+procedure exists — written out in `bandit-nosec-comment-hygiene`'s § Testing
+Strategy — but it lives as **prose inside a Frozen spec**, so when the procedure
+improves the record cannot be corrected. That is the defect this closes.
+
+It earns being a tool for a second reason: the procedure has four traps, and
+each one gave a *wrong answer* the first time it was run by hand. All four fail
+silently — they produce a confident wrong number, not an error.
+
+## The four traps, and where each is encoded
+
+| Trap | What went wrong by hand | Where it lives now |
+| --- | --- | --- |
+| **Rows are not keys** | 362 raw result rows collapse to 239 distinct `(filename, test_id, issue_text)` keys; comparing row counts answers a different question | `scan()` reports both and diffs only the key set |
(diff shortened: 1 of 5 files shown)


eugenelim:
Amended during the merge rebase, both worth flagging:

**1. AC10's ordering resolved.** #977 merged first, so this PR merges second and removes `compare-bandit-suppressions-tool` from `workspace.toml [backlog].open` — the disposition the two PRs agreed on.

**2. The tool found a message bug in itself.** Run for its actual purpose against this batch's own history (`7f186a0b -> main`), it labelled three findings in `tools/export_work_index.py` as `suppression WEAKENED at head`. That file is not in any of these PRs — #971 added it, so those are *new code*, not a weakened suppression. A finding present on one side only has two possible causes and the finding diff cannot distinguish them, so the rows now name both, and the spec gains AC3b. The resolved-id inventory is the half that isolates suppression changes, and it reported the result that matters:

```
resolved suppression ids: identical across 16 shared file(s) (23 directives)
```

**No PR in this batch changed any pre-existing suppression** — the only additions are the new files' own.

PR #1000, opened 2026-08-18
perf(sast): batch the semgrep self-test; archive pip-audit batching as unsafe

eugenelim:
## What does this change?

Batches the semgrep ratchet self-test from five `semgrep` processes into one, making it
3.4–5.0× faster. Also archives an abandoned attempt to do the same for `pip-audit`, and
records eleven pre-existing gaps in the SAST/SCA gate that the investigation surfaced.

## Why?

- Implements: `docs/specs/semgrep-selftest-batching/spec.md` (Shipped)
- Archived, not implemented: `docs/specs/pip-audit-batching/spec.md` (Archived)
- Follows from: ADR-0017 (the SAST/SCA gate), ADR-0086 (`gate-sast` is the critical path,
  so additions to it translate 1:1 into PR latency)

`make sast` is the longest leg of the local gate. The original target was the ten
`pip-audit` invocations; that turned out to be unrecoverable — `pip-audit` resolves every
`-r` file into one environment, so batching trades away per-manifest SCA fidelity, and
upstream transitive upper bounds mean no rule over our own requirements files can prevent
it. The maintainer declined that trade, so the spec is Archived with its measurements and
findings intact rather than deleted.

`semgrep` turned out to be the larger leg anyway (71.2s of 172.0s locally, versus 61.4s
for the collapsible `pip-audit` set), and the self-test half of it was almost pure process
startup — five spawns over five small files, at a ~7.4s startup floor each. Batching that
is safe for precisely the reason batching `pip-audit` was not: semgrep applies
`paths.include`, `--max-target-bytes`, `nosemgrep` and its timeout **per file**, and names
the file in every finding, so nothing merges and attribution is native.

## How do I verify it?

```bash
make sast                                          # exit 0
python3 tools/test-semgrep-argv-boundary.py        # 5/5 passed, one semgrep process
python3 tools/lint-ruff.py
python3 -m pytest tools/test_build_gate_chain.py -q
SKIP_SAST=1 make build-check
```

Timing is deliberately **not** an acceptance criterion — this machine's load average
reached 60 on 10 cores from unrelated work and made absolute figures undependable. The
reported figure is a load-robust interleaved A/B: five invocations at 54.1 / 35.5 / 46.3s
against one at 11.4 / 10.5 / 9.3s, three reps under identical load. On a quiet machine,
29.8s → 8.9s.

The load-bearing check is not that it passes but that it can still **fail**. Seven
mutations, run by patching the loaded module so nothing is left behind in the worktree
(full table in `plan.md` § Verification): a ratcheted target gaining a finding, the rule
neutered, the path key corrupted, an entry dropped from the rule's `paths.include`, every
target missing, an unwired fixture added, and the two skip/missing-rule branches. All
behave as specified.

- **The `pip-audit` invocations.** The point of the branch, and deliberately abandoned —
  see `docs/specs/pip-audit-batching/spec.md` § *Why this was declined*. It is retained
  because re-deriving the measurements and the six verified facts about `pip-audit` 2.10.1
  costs about two hours, and because the obvious approach does not work.
- **Excluding test files from the main semgrep scan.** Worth ~6.6s (20.7% of scan CPU),
  and `bandit.yaml` already excludes test trees — but it is a real coverage reduction, so
  it needs a decision rather than a drive-by.
- **Excluding markdown and JSON from the scan.** They are 58% of the 1801 targets and 2%
  of the time. The obvious-looking win is worth nothing.
- **Vendoring the semgrep registry rules** (~3.8s, permanent maintenance cost, flagged in
  ADR-0017 as a revisit) and **`--jobs` tuning** (already ~5.5× on 10 cores).
- **Four security findings on the control being batched**, all pre-existing and none
  introduced here — most importantly that `FIXED_SCRIPTS` duplicates the rule's
  `paths.include`, so setting it to `[]` exits 0 with the ratchet's production half
  proving nothing. Fixing them would turn a performance change into a redesign of a
  security control, and one of them is ADR-0084-shaped. Surfaced and tracked instead.

## Deferred

Eleven `workspace.toml [backlog].open` entries, none a regression from this PR. Two are
confirmed fail-opens: `sast-ratchet-scope-duplicated-fails-open` (URGENT) and
`sca-requirements-include-lines-unaudited`. The cheapest is `sca-no-strict-flag` —
`pip-audit` without `-S` skips a dependency the advisory service cannot serve, at exit 0,
and adopting `--strict` measured free.

---

## Checklist

- [x] Tests pass locally (`make sast`, `python3 -m pytest tools/test_build_gate_chain.py -q`, both self-tests)
- [x] Lint passes (`python3 tools/lint-ruff.py`); no typecheck surface in this diff
- [x] Self-review run via reviewer subagents — `adversarial-reviewer` and `security-reviewer`, two rounds each; blockers addressed or explicitly surfaced and tracked
- [x] Spec and code agree — both specs updated in this PR; `lint-spec-status.py` exits 0
- [x] Living docs match reality:
  - [x] `docs/product/changelog.md` — not updated: no user-visible behavior change (internal CI/gate tooling only)
  - [x] `guides/` — not updated: no user-facing behavior, config, or interface change
  - [x] `docs/architecture/` — not updated: no material code-structure change
  - [x] `docs/product/roadmap.md` — not updated: completes no roadmap item
- [x] Learnings captured — eleven backlog entries, plus both specs retained as the durable record of what was measured and why the abandoned approach fails


eugenelim:
## Measured on CI

Run [32083483679](https://github.com/eugenelim/agent-ready-repo/actions/runs/32083483679) (this PR) against baseline run 32063058843 (`823cd174`, push to main). Same method both times: filter the log to the `gate-sast` job's `Run make sast` step, timestamp each line, attribute a gap to the line after it.

**The change itself — the semgrep self-test:**

| | before | after | delta |
| --- | ---: | ---: | ---: |
| `test-semgrep-argv-boundary.py` | 8.51s | 1.57s | **−6.94s** (5.4×) |

**Everything else in the step, which this PR does not touch:**

| bucket | before | after | delta |
| --- | ---: | ---: | ---: |
| pip-audit | 75.6s | 65.7s | −9.9s |
| semgrep main scan | 42.3s | 39.7s | −2.6s |
| bandit | 11.0s | 10.1s | −0.9s |
| npm | 1.9s | 1.9s | −0.0s |

Those four moved by −13.4s combined with no code change behind them, so **the step-level 140s → 120s and the job-level `gate-sast` 160s → 144s are mostly run-to-run variance, not this PR.** `gate-main` moved 184s → 151s in the same pair, untouched, which bounds the noise at ±30s or so at job level. The honest attributable saving is the 6.94s above.

**Note the CI and local savings differ by a lot, and the reason is informative.** Locally the self-test went 29.8s → 8.9s (~21s), because semgrep costs ~6s per spawn on that machine. On CI the same five spawns cost 8.5s total — roughly 1.7s each — so there was far less startup to reclaim. The *ratio* is consistent (5.4× on CI, 3.4–5.0× locally); the absolute base was not. An earlier comment in this PR's history inferred CI would save ~21s by analogy with local; that inference was wrong and this measurement supersedes it.

So: a real but modest CI win, a solid local win, and the larger `make sast` costs (pip-audit ~66s, semgrep main scan ~40s) are untouched and remain the levers — see the eleven backlog entries this PR files, and `docs/specs/pip-audit-batching/spec.md` for why the pip-audit one is not safely recoverable.

PR #1042, opened 2026-08-19
feat(journeys): complete priority journey decisions, evidence, and gate navigation

eugenelim:
## What does this change?

Every decision point on the 12 canonical journey pages becomes a shareable, keyboard-reachable link whose address is stable across rewording and reordering. The three priority journeys gain their approved eyebrow and worked transcript, and five new controls hold the state this establishes.

## Why?

Implements `docs/specs/journey-page-completion/spec.md`. 14 of 15 ACs met.

**AC15 is deliberately left unticked.** Its automated clauses pass, but it also requires "no Major design-review issue against the governing principles", which `plan.md` § Approach designates *recorded manual verification*. That is a human sign-off and is not mine to give.

AC15 also dropped "in both approved themes". Journey routes are marketing routes, which `site-browser-quality-gate` AC1 exercises **without theme mutation**, and their emitted pages carry no `data-theme`; that spec's AC2 owns dual-theme coverage for the two `/docs/` routes. No approved width, theme, overflow tolerance, or axe threshold changed — the clause named a condition these routes cannot express. Recorded in the plan changelog.

## How do I verify it?

```bash
rm -rf dist && make build          # dist/ is untracked and goes stale on pull
make build-check                   # includes SAST/SCA
PYTHONPATH=packages/agentbundle:packages/credbroker PYTHONDONTWRITEBYTECODE=1 \
  python3 -m pytest tools/test_journey_editorial_decisions.py -q
npm run test --prefix web
npm run test:e2e:gate --prefix web
```

Measured on this branch: `build-check` exit 0 with SAST/SCA, e2e **127 passed / 0 failed / 0 skipped**, web unit 91, ledger control 3, ci-parity 104/104, rendered links 57,919 across 269 pages clean, journey-contract and web parity 18/18, projection re-sync changes no file.

**Every control is mutation-proved, not assumed.** Reverting either drifted label, deriving fragments from ordinals, disabling the generated-requires-IDs schema rule, forcing the linter's hand-authored key, dropping `decisionGateIds` from a generated journey, altering an eyebrow or one transcript word, inserting a blank line or extra indentation, reducing a Markdown hard break from two spaces to one, adding editorial copy to a non-priority journey, or removing the gate heading's `tabindex="-1"` each fails exactly the control that claims to cover it. Every restoration was verified byte-identical with `cmp`.

Two controls existed only because review kept finding the state unheld:

- The 34 `(journey, id, label)` triples had no automated check. Adding one immediately found **two labels that had never been applied** — `atlassian/review-story-drafts` and `product-strategy/approve-prfaq` — while all 15 browser cases were green. AC2 was false and no gate could see it.
- The priority eyebrow and transcript were verified only by a one-time hand read, though the evidence contract requires exact emission. They are now compared to the ledger line for line. Three review rounds were needed to make that comparison actually support the word "verbatim": `.strip()` hid blank-line and indentation drift, then `rstrip(" ")` tolerated any trailing spaces when a Markdown hard break requires two or more.

- **No pack or plugin version bump.** Verified against the rebuilt 1644-file `dist/`: no `JOURNEY.md` ships and zero installed files carry a semantic gate id, so the migration changes no installed output. AC13's consequence rule did not fire.
- **The six hand-authored journeys.** They keep display-only `yourDecisions` and emit no fragments. The approved ledger has no mappings for them, so inventing IDs would have created a public link contract nobody approved. Deferred as `legacy-hand-authored-journey-gate-mappings`, registered and cross-referenced from the spec.
- **`atlassian`'s pre-existing transcript.** Transcript confinement is a subset check rather than set equality: the ledger's claim is that no journey *gains* one, and equality would have pinned that grandfathered copy in place forever. Narrowing pre-existing copy is outside this spec.
- **The stale `discovery`/`release` IDs in frozen historical specs** (`platform-site`, `phase4-product-docs-rollout`, `ux-writing-rename`, `pack-journeys-phase2b`). Only the *living* template was corrected; frozen artifacts keep their provenance.
- **The inline `hashchange` script's placement.** I predicted a cold-load focus bug from its byte offset preceding its target, and the browser disproved it: focus comes from the HTML fragment-navigation algorithm focusing a `tabindex="-1"` target. Six cold-load cases now pin both mechanisms; no source change was warranted.

### Diff size

1183 insertions across 43 files exceeds the ~400-line guidance. It is genuinely atomic: 11 of those files are generated journey copies that must move with their pack sources or `lint-web-journey-parity` fails, and the ledger test is unusable until registered in all four of its required places.

### Flagged for a maintainer

The spec's `workspace.toml` entry sits in `[work].queue` while `spec.md` reads `Status: Implementing`, so canonical reconciliation reports `impossible_transition` + `unapproved_spec`. It needs `queue` → `active`. I did not hand-edit it: membership changes route through a canonical work-intake transaction, and `work-intake` exposes no such transition. This does not affect any required check.

Spec: docs/specs/journey-page-completion/spec.md

### DIFF
--- a/.github/workflows/build-check.yml
+++ b/.github/workflows/build-check.yml
@@ -250,6 +250,9 @@ jobs:
       - name: pytest make-free gate chains (windows-build-gate-chain)
         run: python -m pytest tools/test_build_gate_chain.py
 
+      - name: pytest journey editorial decisions
+        run: python -m pytest tools/test_journey_editorial_decisions.py
+
       # The generated guides sidebar: sidebar-config.json is gitignored, so
       # these set-equality, label-parity and determinism guards are the only
       # thing standing between a generator regression and a broken published
--- a/Makefile
+++ b/Makefile
@@ -389,7 +389,7 @@ test:
 	$(PYTHON) -m pytest packs/desk-research/tests/skills/desk-research-project-status/ -q
 	$(PYTHON) -m pytest packs/desk-research/tests/skills/desk-research-project-synthesize/ -q
 	$(PYTHON) -m pytest packs/desk-research/tests/skills/devils-advocate/ -q
-	$(PYTHON) -m pytest tools/test_build_gate_chain.py tools/test_catalogue_tooling_rewire.py tools/test_catalogue_tooling_docs.py tools/test_validate_guides.py tools/test_check_guide_index.py tools/test_catalogue_navigation.py tools/test_documentation_entry_links.py tools/test_build_site_link_rewrites.py tools/test_check_rendered_site_links.py tools/test_build_site_routing.py tools/test_check_docs_contrast.py tools/test_build_site_inventory.py tools/test_build_site_projection.py tools/test_build_site_sidebar.py tools/test_browser_gate_subset.py -q
+	$(PYTHON) -m pytest tools/test_build_gate_chain.py tools/test_journey_editorial_decisions.py tools/test_catalogue_tooling_rewire.py tools/test_catalogue_tooling_docs.py tools/test_validate_guides.py tools/test_check_guide_index.py tools/test_catalogue_navigation.py tools/test_documentation_entry_links.py tools/test_build_site_link_rewrites.py tools/test_check_rendered_site_links.py tools/test_build_site_routing.py tools/test_check_docs_contrast.py tools/test_build_site_inventory.py tools/test_build_site_projection.py tools/test_build_site_sidebar.py tools/test_browser_gate_subset.py -q
 	$(PYTHON) -m pytest tools/test_workspace_status.py tools/test_workspace_status_cli.py -q
 	$(PYTHON) -m pytest tools/test_check_artifact_contents.py -q
 	$(PYTHON) -m pytest \
--- a/contracts/catalogue-index.schema.json
+++ b/contracts/catalogue-index.schema.json
@@ -115,7 +115,8 @@
                     "useItWhen": {"type": "string"},
                     "youProvide": {"type": "string"},
                     "youReceive": {"type": "string"},
-                    "yourDecisions": {"type": "array", "items": {"type": "string"}}
+                    "yourDecisions": {"type": "array", "items": {"type": "string"}},
+                    "decisionGateIds": {"type": "array", "items": {"type": "string"}}
                   }
                 }
               }
--- a/docs/product/changelog.md
+++ b/docs/product/changelog.md
@@ -37,6 +37,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Every decision point on a journey page is now a link you can share.** Each
+  place the agent pauses for you has a stable address, so you can send someone
+  straight to the decision itself rather than to the top of the page. The links
+  survive rewording and reordering: if a decision is renamed or moved, an address
+  you shared earlier still lands on the same decision. They also work from the
+  keyboard — tab to a decision, press Enter, and the page moves focus to it.
+- **Decision labels read as plain language.** They describe what you are being
+  asked to do — "Approve the plan", "Confirm the backlog scope" — instead of
+  exposing internal codes. No page shows an internal identifier as visible text.
+- **The core, product engineering, and release engineering journeys open with a
+  one-line summary and a worked example.** Each shows an abbreviated session so
+  you can see the shape of the exchange, and where your decisions fall in it,
+  before committing to the journey.
+
 ### [core][2.9.1] — 2026-08-19
 
 #### Added
--- a/docs/product/changelog.md
+++ b/docs/product/changelog.md
@@ -37,6 +37,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Journey authors can record which decisions a reader meets, and in what
+  order.** A pack's `JOURNEY.md` may now carry an optional
+  `contract.decisionGateIds` — an ordered list of `humanGates[].id` values. It
+  holds identifiers only; the wording readers see still comes from each gate's
+  label. The field is optional and `yourDecisions` remains required, so every
+  pack authored before this stays valid with no edit.
+
 ### Changed
 
 - **Every decision point on a journey page is now a link you can share.** Each
--- a/docs/specs/journey-page-completion/plan.md
+++ b/docs/specs/journey-page-completion/plan.md
@@ -316,3 +316,23 @@ no version, dependency, migration, or infrastructure change.
   normalised rather than compared. Proved by MP16 (hard break 2 -> 1 space now
   fails, previously passed) and MP17 (2 -> 3 spaces still passes, since three
   spaces is still a hard break).
+- 2026-08-19: extended the published catalogue contract additively, under
+  explicit authorization, after CI surfaced a conflict local gates could not see.
+  `tests/roster/test_catalogue_wave4_live_contracts.py` arrived in #1025 — one of
+  the eleven commits this branch rebased onto — and is wired only in
+  `build-check.yml`, never the Makefile, so `make build-check` passed locally
+  while `gate-main` failed. Its validator treats `contract.*` as a closed set and
+  also requires `yourDecisions`, so the ratified `contract.decisionGateIds` was
+  rejected on all 12 canonical packs. Because the approved ledger fixes the field
+  inside `contract`, relocating it was not available, even though unknown
+  top-level keys are accepted today. AC13 routes a discovered installed-output
+  change to a spec amendment rather than a silent version bump, so the work
+  stopped and asked. The authorized resolution is additive: `decisionGateIds`
+  optional in `journey_validator`, both byte-identical schema copies, and both
+  authoring-standard copies; `yourDecisions` remains required and was restored
+  byte-for-byte from the merge-base into all 12 packs. Adopter impact is nil.
+  MP18-MP20 prove the change did not open the gate: removing `yourDecisions`
+  still fails, a non-array `decisionGateIds` fails, and an unknown contract field
+  still fails. The scaffold copy is a projection, so `manifest.json` was
+  regenerated with `tools/catalogue/sync_authoring_scaffold.py --write` rather
+  than hand-edited.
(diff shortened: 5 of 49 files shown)


eugenelim:
## Blocked: `gate-main` fails on a published-contract conflict I should not resolve unilaterally

`gate-main` and `make build-check` are red. Both trace to **one** failure — `make build-check` only aggregates `gate-main`:

```
tests/roster/test_catalogue_wave4_live_contracts.py::test_live_catalogue_indexes_every_manifest_pack
CatalogueIndexError: unknown contract field: decisionGateIds
```

### Why local was green

That test arrived in **75bfaf9c (#1025, catalogue semantic contracts index)** — one of the 11 commits this branch rebased onto. It did not exist at the merge-base (4b89ed31). It is wired only at `.github/workflows/build-check.yml:451` and appears nowhere in the `Makefile`, so local `make build-check` passes while CI fails. Local post-rebase state is otherwise fully green: build-check exit 0, 4-stage site build 0, browser gate 127 passed / 0 failed, links 57,966 across 269 pages clean, tree clean.

### The conflict

`agentbundle.catalogue_tooling.journey_validator` treats `contract.*` as a **closed** set:

- `CONTRACT_KEYS = ("useItWhen", "youProvide", "youReceive", "yourDecisions")` rejects any unknown `contract` field, so `contract.decisionGateIds` fails on all 12 canonical packs.
- It also **requires** `yourDecisions`; allowing the new field alone yields `missing required contract key: yourDecisions`, then a `KeyError` at `journey_validator.py:74`.

That contract is published in five places, two pairs of which CI requires to stay identical:

| File | Role |
|---|---|
| `contracts/catalogue-index.schema.json` | public contract |
| `packages/agentbundle/agentbundle/_data/catalogue-index.schema.json` | bundled copy — a test asserts byte-identity with the above |
| `guides/_shared/reference/catalogue-authoring-standards.md` | adopter-facing authoring docs |
| `.../catalogue-scaffold/guides/_shared/reference/catalogue-authoring-standards.md` | scaffold copy — a test asserts its `## Journey format` section is identical |
| `journey_validator.py` | the enforcer |

The approved ledger is explicit that the field belongs there — "`contract.decisionGateIds` contains only the ordered internal IDs" — so relocating it to top-level frontmatter is not available to me, even though unknown *top-level* keys are accepted today (`humanGates`, `eyebrow`, `goodOutputDescription` and seven others already pass).

### Why I stopped rather than fixed it

This is AC13's stated trigger: *"if an installed functional behavior change is discovered, implementation stops for a spec amendment rather than silently versioning it."* Widening agentbundle's published catalogue contract — and relaxing one of its required fields — is an adopter-facing interface change that was not in this spec's approved scope, because the enforcer did not exist when the spec was approved. `Gate G — agentbundle release impact` exists to police exactly this.

Not merging: required checks are red, and I will not suppress or loosen a test to turn them green.

Awaiting a decision on scope. Details and options are with the requester.

eugenelim:
## Resolved — all five required contexts green

Two separate blockers surfaced, both from CI-only controls that local `make build-check` cannot see.

### 1. The closed contract set

`journey_validator` treats `contract.*` as closed and also requires `yourDecisions`, so the ratified `contract.decisionGateIds` was rejected on all 12 canonical packs. The approved ledger fixes the field inside `contract`, so relocating it was not available.

Resolved **additively**, on the requester's decision after the conflict was surfaced under AC13:

- `decisionGateIds` is optional; `yourDecisions` stays required and was restored byte-for-byte from the merge-base into all 12 packs (verified identical 12/12).
- The unknown-field check now tests the union of required and optional keys, so the set stays closed to everything else. A new type check requires an array of strings.
- Both byte-identical schema copies gained the property; neither `required` array changed. Both authoring-standard copies document it. The scaffold copy is a projection, so its digest manifest was regenerated with `tools/catalogue/sync_authoring_scaffold.py --write`.

The two keys now coexist by design — IDs drive fragments and ordering, strings remain adopter-facing prose. The renderer prefers the IDs, so nothing renders twice: 34 chip links still match 34 targets, the display-string fallback appears on the six hand-authored journeys and no canonical one, and `Merge the PR` (a decision string that is no gate's label) appears on zero emitted pages.

**No version was bumped.** `docs/product/changelog.md` is itself a Gate G release indicator, so `has_release_indicator` was already satisfied — which is what AC13 asks for: surface the change, don't silently version it.

### 2. The protected-tree guard

`lint-catalogue-curation-guard` hard-fails any changeset touching `packages/agentbundle/**` without an `Engine-Change-RFC:` trailer (RFC-0059 D6). The carve-outs — `build/recipes/**` and `**/tests/**` — don't cover these files.

The changeset now carries `Engine-Change-RFC: RFC-0076`. Stated plainly, because a governance trailer deserves it: RFC-0076 is **Accepted**, owns `contracts/catalogue-index.schema.json` ("Contract location: `contracts/catalogue-index.schema.json` (new, Wave 4)") and the Wave 4 JOURNEY.md contracts, and **#1025 cited this same RFC for a change to this same tree and contract**. It supplies the authorized engine-change path and the visibility the guard exists to create. It does not by itself pre-authorize this particular field — that was the requester's additive-only decision, recorded as a dated amendment in the spec's Boundaries.

I confirmed the guard discriminates rather than merely passing: it fails without the trailer and passes with it.

### Verification

Required contexts: `make build-check`, `gate-main`, `gate-sast`, `gate-export-boundary`, `gate-credbroker` — all pass. `build` (pages.yml) passes, which is the browser gate on the runner.

Locally on this tree: build-check exit 0 with SAST/SCA, browser gate **127 passed / 0 failed**, web unit 91, ledger control 3, journey-contract lint 18 conform with a 9-case self-test, ci-parity 104/104, links 57,966 across 269 pages clean, the CI-only roster test 3 passed, scaffold projection and contract parity green.

Eight mutation proofs cover the new surface, each restored byte-identically. Two of them caught their own errors first: one mutated `packs/` when the linter reads `web/src/content/journeys/`, and the new self-test case silently never ran until it was registered in `main()`'s explicit list — 8 tests became 9.

AC15 remains unticked: its automated clauses pass, but its design-review clause is recorded manual verification awaiting human sign-off.

eugenelim:
## Recorded design review — AC15's final clause: NOT satisfied

AC15 requires "no Major design-review issue against the governing principles", and `plan.md` § Approach designates that clause **recorded manual verification**. This is that record.

**Reviewer:** independent `experience-reviewer` pass (read-only), against `docs/design/principles/tech-site.md` (`status: active`, `applies_to` includes `journeys`) and the shared `quality-floor` checklist. Evidence: 21 rendered screenshots of the three priority routes at 360/375/390/414/1440 with no theme mutation, including the decision section with its first chip reached by keyboard Tab, plus the three source components.

**Verdict: SHIP WITH CHANGES — 3 Major, 0 Blocker.** On the bar as written, AC15's final clause cannot be ticked.

### Majors, each verified independently by arithmetic or the emitted artifact

| # | Finding | Measured | Floor |
|---|---|---|---|
| 1 | `.decision-chip:hover, :focus-visible` sets `--ds-hero-fg` (`#ffffff`) on `--ds-accent` (`#e8952b`) | **2.40:1** | 4.5:1 for 13px/12px weight-400 |
| 2 | `.gate-card__label:focus-visible` / `:target` ring is `--ds-accent` on `#fafaf9` | **2.29:1** | 3.0:1 non-text (WCAG 1.4.11) |
| 3 | `goodOutputDescription` is interpolated into `<p class="good-output">`, so Astro escapes it | literal `**You:**` and backticks in emitted HTML | — |

**1** is quality-floor perceivable contrast plus principle 4 ("shared identity never overrides readability, accessibility"). The token comments condemn it: `--ds-hero-fg` is "primary text **on dark**" and `--ds-accent` is "CTA fill **on dark**", both used here on a light surface. The in-system fix — dark-on-amber, the pairing already defined for CTAs — measures **8.25:1**. Acquiring focus currently makes the chip *less* legible than its unfocused neighbours.

**2** is the indicator a keyboard user relies on *after* activating a chip, once focus has left the chip. The asymmetry is the tell: `.decision-chip:focus-visible` already overrides the global amber ring with `--ds-on-surface`; the same fix was not carried to the destination fifteen lines away.

**3** is principle 2 — evidence beside claims. Verified directly against `build/journeys/release-engineering/index.html`. Three consequences from one cause: markup characters ship as adopter copy, HTML whitespace collapsing destroys the speaker-turn structure that makes a transcript verifiable, and the page's one concrete artifact is visually indistinguishable from the marketing prose above it.

### Attribution

All three are live because of this PR. Checked against merge-base `4b89ed31`: `.decision-chip:hover` did not exist, and none of the three priority journeys carried a `goodOutputDescription`. So the failing hover pairing is new here, and the transcripts are what expose the pre-existing renderer's escaping.

### Why the automated gates did not catch 1 and 3

- **Axe cannot express finding 1.** It scans the resting DOM, so a `:hover`/`:focus-visible` declaration never applies during the scan. The browser gate this PR extended is structurally blind to state-dependent contrast.
- **The ledger control checks content fidelity, not presentation.** It proves the transcript matches the approved ledger byte-for-byte — which it does. AC8's "verbatim" holds at the source while the rendered page shows asterisks. There was no control on the emitted form.

### Notable lower-severity items

- **Advisory:** the *resting* chip passes at **4.59:1** against a 4.5 requirement — a 2% margin held up by the fill being only 10% opaque (`--ds-accent-subtle: #e8952b1a`), so it fails silently on any darker surface with no axe signal. Worth pinning with an assertion on the composited pair.
- **Minor:** `"Approve the G4 handoff — merge the PR"` sits in `web/src/content/journeys/iac-terraform.md`'s `yourDecisions`, a field the schema calls adopter-facing. Invisible today only because the renderer prefers `decisionGateIds`.
- **Minor:** `Decision 1 of 1` on `release-engineering` is an ordinal doing no work; suppress it when the count is one.
- **Minor:** one `docsUrl` destination carries two labels on one page ("Read the reference" / "Read the full reference") — principle 3.

### What the review confirmed as working

Decision navigation genuinely leads with the reader's job; no `globalGate` or legacy `G…` code reaches the reader on any of the three routes; the `:target` + `:focus-visible` pairing correctly covers cold fragment loads where programmatic focus will not match `:focus-visible`; the schema's `superRefine` makes the chip set structurally unable to lie about the gate set; and reduced-motion is honoured.

### Disclosure

I seeded the reviewer with a false premise — I had mis-measured the *resting* chip at 2.37:1 by treating a 10%-alpha token as an opaque fill instead of compositing it. I retracted it mid-review. The reviewer independently re-derived 4.59:1, declined to raise a resting-chip finding, and located the real defect in the focused state instead.

**Next:** a follow-up fix PR. Findings 1 and 2 are mechanical token corrections. Finding 3 is a presentation decision on approved editorial content and needs an owner call. AC15 stays unticked until then.

PR #1045, opened 2026-08-19
fix(journeys): repair the three Major design-review findings on journey pages

eugenelim:
## What does this change?

Fixes the three Major findings from the recorded design review of the three priority journey pages — two contrast failures and a transcript that shipped as escaped Markdown — and adds the controls that would have caught each. All three were live because of #1042.

## Why?

`docs/specs/journey-page-completion/spec.md` AC15 requires "no Major design-review issue against the governing principles". The recorded review (on #1042) returned **SHIP WITH CHANGES** with three Majors, so the clause could not be ticked.

| # | Defect | Was | Now | Floor |
|---|---|---|---|---|
| 1 | Chip `:hover`/`:focus-visible` text | **2.40:1** (white on amber) | **8.07:1** | 4.5:1 |
| 2 | Gate heading `:target`/`:focus-visible` ring | **2.29:1** page / **2.08:1** card | **16.74:1** | 3.0:1 (WCAG 1.4.11) |
| 3 | Transcript | literal `**` and backticks, turns collapsed into one paragraph | 6/9/5 attributed turns, zero markup characters | — |

On **1**, both token comments read "on dark" — `--ds-hero-fg` is "primary text on dark", `--ds-accent` is "CTA fill on dark" — and they were paired on a light surface. `--prim-amber-900` is annotated for exactly this job, but components reference zero `--prim-*` tokens directly, so `--ds-cta-primary-fg` is the pairing that fits the convention.

On **2**, this is the indicator a keyboard user relies on *after* activating a chip, once focus has left the chip and its own ring is gone. `.decision-chip:focus-visible` had already overridden the global amber ring; the same fix simply had not been carried fifteen lines away. One interaction now speaks one focus language.

On **3**, `parseTranscript` returns `{speaker, parts}` and the template renders real elements in the page's mono register. No Markdown dependency was added and there is no `set:html`, so no injection surface. Soft-wrapped source lines rejoin into one utterance. **The approved wording is untouched — only its presentation changed.**

## How do I verify it?

```bash
rm -rf dist && make build
make build-check
npm run test --prefix web            # 103 passed, incl. 12 parser tests
npm run test:e2e:gate --prefix web   # 136 passed
```

Measured on this branch: build-check exit 0 with SAST/SCA, browser gate **136 passed / 0 failed** (was 127: +6 focus-state contrast, +3 transcript), web unit **103** (was 91), ledger control 3, journey-contract lint 18 conform with a 9-case self-test, ci-parity 104/104, links 57,973 across 269 pages clean, CI-only roster test 3 passed, 4-stage build all exit 0.

**Why the existing gates missed these.** Axe scans the resting DOM, so a `:hover`/`:focus-visible` declaration never applies during the scan — that is how a 2.40:1 chip passed a green accessibility gate. And the ledger control compares source to source; it was green throughout while the rendered page showed asterisks. `expectTextContrast` / `expectOutlineContrast` now enter the state first, then measure, and they composite alpha layers rather than treating a translucent token as opaque — `--ds-accent-subtle` is `#e8952b1a`, whose trailing `1a` is 10% alpha, and treating that as an opaque fill reports 2.37:1 for a chip that renders at 4.59:1.

**MP24–MP26** revert each fix and reproduce 2.40:1, 2.08:1 and a missing transcript; all restorations verified byte-identical with `cmp`.

- **A Markdown renderer for finding 3.** "Add no dependency" is ratified, so the parser handles the two constructs the approved transcripts actually contain — a leading `**Speaker:**` label and inline `` `code` `` — verified by enumerating them. An unpaired backtick stays literal rather than swallowing copy.
- **`set:html`.** It would have been shorter and would have introduced an injection surface for no benefit.
- **A two-column speaker/utterance grid.** This page carries no media query, and a speaker as long as "Independent reviewer" squeezes the utterance at 360px. Turns stack instead.
- **The decorative `border-left: 4px solid var(--ds-accent)` on the gate card.** It conveys no state, so WCAG 1.4.11 does not reach it, and changing it would be scope creep.
- **The review's Minors and Advisories** — deferred by owner decision to keep this to the three Majors: suppressing `Decision 1 of 1`, the two labels on one `docsUrl` destination, a persistent affordance distinguishing navigational chips from label chips, and a latent `"Approve the G4 handoff"` string in `iac-terraform.md`'s adopter-facing `yourDecisions` (invisible today only because the renderer prefers `decisionGateIds`).

## Lifecycle

Moves the spec's `workspace.toml` entry from `queue` into `["ini-002".work]`'s existing empty `active` list. A merged artifact reading `Status: Implementing` inside a queue made canonical reconciliation report `impossible_transition` and `unapproved_spec`; it now reports `classification: active` with no blocking needs. It moves to `shipped` when AC15 is ticked. `site-shared-chrome` stays queued and untouched.

**AC15 remains unticked pending re-review of these fixes.** I will post the re-review verdict on this PR.

Spec: docs/specs/journey-page-completion/spec.md

### DIFF
--- a/docs/product/changelog.md
+++ b/docs/product/changelog.md
@@ -48,6 +48,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **The example session on a journey page now reads as a session.** Each turn is
+  attributed to whoever spoke it and sits on its own line in a terminal-style
+  register, instead of running together in one paragraph with stray asterisks and
+  backticks showing. The wording is unchanged.
+- **Keyboard focus on a journey decision is easier to see.** Focusing a decision
+  no longer makes its label harder to read than the ones beside it, and the
+  heading you land on after activating one now takes the same clearly visible
+  outline the decision itself uses.
 - **Every decision point on a journey page is now a link you can share.** Each
   place the agent pauses for you has a stable address, so you can send someone
   straight to the decision itself rather than to the top of the page. The links
--- a/docs/specs/journey-page-completion/plan.md
+++ b/docs/specs/journey-page-completion/plan.md
@@ -356,3 +356,30 @@ no version, dependency, migration, or infrastructure change.
   Also moved the spec's workspace entry from `queue` to the initiative's existing
   empty `active` list, clearing the `impossible_transition` and `unapproved_spec`
   findings that a merged-but-Implementing artifact produced in a queue.
+- 2026-08-19: re-review of the three Major fixes confirmed all three resolved and
+  found two more. One was a regression this branch introduced: the transcript fix
+  changed a shared template, and `atlassian` ships a `goodOutputDescription` that
+  is prose rather than a session — the grandfathered field whose allowlist entry
+  this spec already carries. It was rendering as a single unattributed turn: an
+  empty `<dt>` around 127 words of author prose in the 13px mono session register,
+  asserting session evidence for a paraphrase. The template now branches on the
+  parse result and prose renders in the prose register. The control could not see
+  it because it looped only the three priority routes, so it is now enumerated
+  from the built site and covers all 18 journey routes; MP27 forces every route
+  into the session register and fails exactly one of eighteen. A guard case
+  asserts the enumeration is non-empty, because an empty route list would make
+  every case vacuous.
+  The second was a Minor this branch made conspicuous rather than caused: the gate
+  heading and its card both painted an identical 3px near-black ring at the same
+  offset, verified from computed styles on both elements. Invisible while both sat
+  at 2.08:1; obvious at 15:1. The heading now owns the indicator for both
+  `:focus-visible` and `:target`, so the cold-load path that needs its own
+  selector survives — programmatic focus on a `tabindex="-1"` heading matches
+  `:focus-visible` only by Chromium heuristic — and the card computes
+  `outline-style: none`.
+  Deferred to its own PR by owner decision: the review's M2, the global
+  `:focus-visible` amber ring in `web/src/styles/base.css`, which measures 2.29:1
+  on light surfaces against a 3:1 non-text floor and 8.07:1 on the dark hero where
+  it is correct. It is pre-existing, owned by the design-system layer, and lands on
+  every marketing route, so it warrants its own review rather than riding along in
+  a journey-pages fix. AC15 stays unticked until it is resolved and re-reviewed.
--- a/docs/specs/journey-page-completion/plan.md
+++ b/docs/specs/journey-page-completion/plan.md
@@ -336,3 +336,23 @@ no version, dependency, migration, or infrastructure change.
   still fails. The scaffold copy is a projection, so `manifest.json` was
   regenerated with `tools/catalogue/sync_authoring_scaffold.py --write` rather
   than hand-edited.
+- 2026-08-19: fixed the recorded design review's three Major findings. Two were
+  contrast failures invisible to the browser gate by construction: axe scans the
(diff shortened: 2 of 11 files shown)


eugenelim:
## Recorded re-review — AC15's final clause: still NOT satisfied

Second independent `experience-reviewer` pass, against `docs/design/principles/tech-site.md` and the shared `quality-floor`, with before/after screenshots of the three priority routes at all five approved widths plus keyboard-focused decision sections.

**Verdict: SHIP WITH CHANGES.** All three original Majors confirmed resolved — the reviewer re-derived every contrast figure independently rather than taking mine (chip **8.09:1**, ring **16.49:1** page / 14.99:1 card, and zero Markdown characters in the built HTML). It then found two Majors, one of which this branch introduced.

### Fixed in this PR

**M1 — regression, introduced by the transcript fix.** `parseTranscript` had no designed output for a `goodOutputDescription` carrying no `**Speaker:**` labels, which the spec explicitly permits and `atlassian` ships. It emitted `<dt class="transcript__speaker"></dt>` around a 127-word author paraphrase in the 13px mono session register — asserting captured-session evidence for a summary, the inverse of the principle the work was serving. **And the new control could not see it:** the non-empty-speaker assertion looped only the three priority routes, so the one route that violated it was outside the loop. The register is now chosen by content shape, and the suite enumerates journey routes from the built site — **18 routes, up from 3** — with a guard case asserting the enumeration is non-empty, because an empty list would make every case vacuous. MP27 forces every route into the session register and fails exactly 1 of 18.

**m3 — Minor, made conspicuous by the contrast fix.** The gate heading and its containing card both painted an identical `3px solid` near-black outline at the same offset; I confirmed it from computed styles on both elements, since an element-clipped screenshot cannot show a ring that `outline-offset` draws outside the box. Invisible while both sat at 2.08:1, conspicuous once they cleared 15:1. The heading now owns the indicator for `:focus-visible` and `:target` both — the cold-load path still needs its own selector, because programmatic focus on a `tabindex="-1"` heading matches `:focus-visible` only by Chromium heuristic — and the card computes `outline-style: none`.

**m4** also taken, since it was the same lines: `<ol>` replaces `<dl>`, which the HTML standard calls inappropriate for dialogue.

### Deferred by owner decision

**M2 — pre-existing, site-wide.** `web/src/styles/base.css` sets the global `:focus-visible` ring to `2px solid var(--ds-accent)`. That measures **2.29:1** on the light page and **2.08:1** on the card, under the 3:1 non-text floor, and lands on `.gate-card__summary`, three `.next-link`s and `.prereqs__list a`. On the dark hero it is 8.07:1 and correct. Options measured: `--ds-accent-deep` **5.43:1** keeps the amber character; `--ds-on-surface` **16.49:1** matches what the chip and gate now use.

It is deferred to its own PR because it is pre-existing, owned by the design-system layer (`site-design-system-spec` and `site-ui-primitives`, both Shipped), and changes focus appearance on every marketing route — so it warrants review across all eight rather than riding along in a journey-pages fix.

Also deferred: the review's remaining Minors and Advisories — inline code and speaker attribution sharing one treatment, suppressing `Decision 1 of 1`, two labels on one `docsUrl`, a mono-specific measure for the transcript, two literal `font-weight: 600` values where a token exists, and a latent `"Approve the G4 handoff"` string in `iac-terraform.md`'s adopter-facing `yourDecisions`.

### Disclosure

In the first review I planted a false premise — I had mis-measured the *resting* chip at 2.37:1 by treating a 10%-alpha token as an opaque fill — and retracted it mid-review. The reviewer independently re-derived 4.59:1, declined to raise a resting-chip finding, and located the real defect in the focused state instead.

**AC15's final clause cannot be ticked until M2 is resolved and re-reviewed.**

PR #1048, opened 2026-08-19
chore(workspace): shape round 13 and record the glob-vs-MEMBERS finding

eugenelim:
Round 13 is not commissioned; the shaping item is the commissioning decision
itself, carrying three candidate arms that each trace to an existing residual and
the ordering rationale for why the group-split-cost arm goes first. The addon
arm is marked blocked on round 12's confinedRemove fix because it feeds that
helper the input shape that trips the defect.

The glob finding is recorded as a residual rather than fixed here: two tools
derive a fact about promoted artifacts by globbing the evidence tree instead of
iterating MEMBERS, so the selector does not match the set the claim is about.

### DIFF
--- a/workspace.toml
+++ b/workspace.toml
@@ -57,6 +57,18 @@ backlog = [
   # because real surfaces have opposite needs. Should destination become the one
   # place a session's per-destination policy lives (egress + worker + more)?
   {slug = "rfc0088-destination-scoped-policy-axis", type = "shape"},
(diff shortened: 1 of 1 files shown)


eugenelim:
Closing unmerged — round 12 absorbs these instead of the workspace carrying two more entries. Branch retained on the remote if the shaping item is wanted later.

PR #1052, opened 2026-08-19
feat(tools): worktree hygiene — scan/clean doctor, leased preview ports, selective bootstrap

eugenelim:
Two tasks against `docs/specs/worktree-runtime-hygiene/spec.md`.

## Why

Nine parallel worktrees share one machine. **Concurrency, not disk, is the problem** — measured before building anything:

| | |
|---|---|
| All worktrees combined | ~2% of a 402 GB volume |
| Playwright browsers | **already shared once** — one build per browser, revision pinned by `playwright-core` 1.62.0 |
| Genuinely duplicated | only `node_modules` — 569 MB × 4 bootstrapped worktrees |

So this ships measurement, a conservative deletion boundary, and collision-free concurrency — not a reclamation sweep. Byte accounting is diagnostic output.

## Task 1 — `worktree_hygiene.py` (`51135b88`, `e6a8f594`)

`scan` classifies bytes per worktree and labels whether it measured **allocated or logical** bytes, since sparse files and APFS clones diverge by >2× and an unlabelled total is a false claim.

`clean` is dry-run by default and **confined to the current worktree** — "every worktree" is deliberately inexpressible, making the cross-worktree accident *unrepresentable*. Real deletion needs `--apply` + a category; dependencies need `--include-dependencies`.

Refuses: symlink / junction / mount-point / filesystem-boundary escapes, the git common dir and nested `.git`, tracked files, non-ignored paths, `.loop-run` / `.context`, locks, protected worktrees, and paths an installed distribution resolves into.

## Task 2 — `frontend_runtime.py` + `bootstrap.py` (`7841aa80`)

`PREVIEW_PORT` reads `ARR_PREVIEW_PORT`, default 4321, **throwing** on invalid rather than falling back — a silent fallback reintroduces the collision invisibly. `playwright.config.ts` untouched; it already derives everything from that constant.

`gate` leases a port through the **Git common directory**, the one location every linked worktree shares. Proven: three concurrent wrappers take 4321 / 4322 / 4323 and release cleanly.

`bootstrap` installs exactly one profile — `web`, `docs-site`, or `sites`. No implicit "all", so a governance worktree never installs both trees.

## Three bugs that only running it could find

1. `git check-ignore` exits **1** to mean "nothing ignored" — success, not failure.
2. `git check-ignore` **rejects `--literal-pathspecs`**; only `ls-files` accepts it. A review round certified the broken version as correct; one real-repo run caught it.
3. Signalling the child tree from inside the handler **deadlocked the wrapper** — `Popen.wait()` isn't reentrant, and the handler ran on top of the main thread's own wait. The tree died and the process hung, so the lease never released. Handlers now only send.

## Verification

`ruff` · `mypy` · `bandit` (medium/medium) · **81 tests** · `tools/` 831 passed · `lint-spec-status` 0 · `lint-traceability` 0 · roster validation 7 passed (= `main` baseline)

**19 guards mutation-proven** — each disabled independently, suite goes red, tree restored byte-identical.

Real-machine exercise: 13 worktrees scanned, tracked-directory guard fired on genuinely tracked `build/` dirs, three concurrent gates got distinct ports, SIGTERM killed the grandchild and released the lease in 0.01s.

## Deferred, recorded rather than papered over

- A mutator racing the final re-assertion needs a cooperative lease.
- The probe closes its socket before Astro binds, so an unrelated machine-local listener can still take the port in between. Participating wrappers are protected; other processes aren't.
- **No real cleanup performed** — every worktree was in use, which the spec names as a valid outcome.
- The pinned gate surface (`test:e2e:gate`, `pages.yml`) is deliberately untouched; `tools/test-pages-workflow.py` fixes both.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,12 @@ build/
 # build/** override so it applies inside the exception too).
 __pycache__/
 .pytest_cache/
+# Type- and lint-checker caches. Unignored, these are staged by `git add -A`
+# (~62MB of .mypy_cache across worktrees), and tools/repo/worktree_hygiene.py
+# only ever deletes ignored paths, so its --test-artifacts category is inert
+# without them.
+.mypy_cache/
+.ruff_cache/
 
 # Build pipeline output.
 dist/
--- a/Makefile
+++ b/Makefile
@@ -396,6 +396,8 @@ test:
 	$(PYTHON) -m pytest tools/test_build_gate_chain.py tools/test_journey_editorial_decisions.py tools/test_catalogue_tooling_rewire.py tools/test_catalogue_tooling_docs.py tools/test_validate_guides.py tools/test_check_guide_index.py tools/test_catalogue_navigation.py tools/test_documentation_entry_links.py tools/test_build_site_link_rewrites.py tools/test_check_rendered_site_links.py tools/test_build_site_routing.py tools/test_check_docs_contrast.py tools/test_build_site_inventory.py tools/test_build_site_projection.py tools/test_build_site_sidebar.py tools/test_browser_gate_subset.py -q
 	$(PYTHON) -m pytest tools/test_workspace_status.py tools/test_workspace_status_cli.py -q
 	$(PYTHON) -m pytest tools/test_worktree_hygiene.py -q
+	$(PYTHON) -m pytest tools/test_frontend_runtime.py -q
+	$(PYTHON) -m pytest tools/test_bootstrap.py -q
 	$(PYTHON) -m pytest tools/test_check_artifact_contents.py -q
 	$(PYTHON) -m pytest \
 		tools/test_lint_agents_md_diataxis_block.py \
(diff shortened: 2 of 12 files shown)


eugenelim:
## RFC-0090 change-sizing declaration

RFC-0090 landed on `main` (`de10c9e2`) after this work started, and this PR is now governed by it. Reporting the measurements the policy asks for, rather than leaving the size unexamined.

**Volume** (vs `origin/main`, added lines):

| File | + |
|---|---|
| `tools/repo/worktree_hygiene.py` | 1227 |
| `tools/test_worktree_hygiene.py` | 920 |
| `tools/repo/frontend_runtime.py` | 640 |
| `tools/test_frontend_runtime.py` | 540 |
| `tools/test_bootstrap.py` | 129 |
| `tools/repo/bootstrap.py` | 102 |
| `web/src/test/site-base.test.ts` | 39 |
| `Makefile` | 20 |
| `web/src/test/e2e/site-base.ts` | 18 |
| *(prose, excluded from sizing)* `spec.md` 122, `plan.md` 26 | — |
| *(config)* `.gitignore` | 6 |

- **Raw diff lines:** 3789 (triage only)
- **Reviewable behavior + test lines:** 3635 — **tail-triage trigger tripped** (>2000)
- **Median reviewable lines per changed file** (prose and generated output excluded, 9 files): **129**
- **Tail shape: MIXED** (>60, <200)

**Decomposition.** MIXED requires dependency-ordered working layers. This PR is a stack of three curated, independently testable commits, not a single blob:

1. `ed4e25b9` — the scan/clean doctor. **Independently verified:** this commit was pushed alone and passed 33/33 CI as a standalone PR before any task-2 code existed, so the layer demonstrably leaves the repository working.
2. `fba421d5` — bandit B108 fix for that layer.
3. `b3d05264` — leased ports, browser-cache resolver, bootstrap profiles. Touches disjoint files from layer 1 (`frontend_runtime.py`, `bootstrap.py`, `site-base.ts`); adds one Makefile block and one spec section.

Each layer is separately reviewable, and the second cannot break the first: no file is edited by both except `Makefile` (append-only, different hunks) and `spec.md`.

**Not claiming a no-split exemption.** A split into two PRs would also have preserved working layers — layer 1 proved that. The layers are presented as a reviewable stack rather than two PRs because the review history (four independent adversarial rounds, 19 mutation proofs) attaches to the pair. If a reviewer would rather see this as two PRs, say so and I will split it; the commit boundaries already permit that with no rework.

PR #1154, opened 2026-08-28
build(deps): land astro 7.2.4, mermaid 11.17.0, vitest 4.1.11 (supersedes #1112, #1113)

eugenelim:
Lands the two open Dependabot group updates as one change, plus one declared
bundled fix. Supersedes #1112 and #1113.

| PR | Head SHA | Directory | Bump |
| --- | --- | --- | --- |
| #1112 | `6c87ccaecda840e0daa4b60a73c73e4036e042d4` | `web/` | `astro` 7.2.2→7.2.4, `vitest` 4.1.10→4.1.11, `@vitest/ui` 4.1.10→4.1.11 |
| #1113 | `046801b88c23e1ae4e58ad698b92aa5644374b34` | `docs-site/` | `astro` 7.2.2→7.2.4, `@astrojs/markdown-remark` 7.2.2→7.2.4, `mermaid` 11.16.1→11.17.0 |

Both commits are cherry-picked unmodified. `git diff <head-sha> HEAD` over each
directory's `package.json` + `package-lock.json` is **empty**, so the landed
bytes are provably the bytes CI already went green on. (Positive control: the
same pathspecs against the branch point yield 706 and 461 diff lines, proving the
pathspecs resolve — "empty output" alone is also what a mistyped pathspec
produces.)

**Why one PR, not two.** Both site builds compose into one `build/` tree and
`web/src/test/rendered-output.test.ts` asserts against the emitted HTML of both,
so landing one alone would leave the published site built by two astro minors and
make that suite's verdict ambiguous. Independently, `docs-site`'s
`@astrojs/markdown-remark` pin must equal astro's exact optional-peer version, so
astro and that package must move together within #1113.

## What did you change, and why?

Six version strings and two lockfiles. Plus one line of
`docs/architecture/reference.md` — see *Bundled fix* below.

### Impact assessment — measured from the lockfiles, not the release notes

**1. `web/`'s markdown renderer moves: satteri 0.9.5 → 0.10.5.**
`web/astro.config.ts` sets no `markdown` config, so it renders through astro's
default processor, which is Sätteri. `web/AGENTS.md` states the split outright:
the two co-deployed sites are deliberately on different engines. The
HTML-parsing path is *replaced*, not churned — `hast-util-from-html`,
`hast-util-from-parse5`, `hast-util-parse-selector`, `hastscript`,
`vfile-location`, and `web-namespaces` are removed, while `parse5` and `entities`
lose a duplicate resolution each. 42 markdown files render through it (22 packs +
20 journeys, the only two globs in `web/src/content.config.ts`).

**2. `docs-site/`'s markdown renderer also moves:** `@astrojs/markdown-remark`
7.2.2 → 7.2.4 and `@astrojs/mdx` 7.0.5 → 7.0.7. `docs-site/astro.config.ts`
imports `unified` *from* `@astrojs/markdown-remark` and builds
`markdown.processor` from it, so the package supplying this site's processor is
itself bumped. The pin gate proves version *agreement*, not output stability.

**3. `docs-site/` ends up with two Sätteri minors at once.**
`@astrojs/starlight@0.41.7` declares `satteri: ^0.9.1` and is not in #1113's
bump set, so the lockfile retains hoisted `satteri` 0.9.5 *and* adds nested
0.10.5 under `@astrojs/markdown-satteri` 0.3.7 — 12 → 21 `@bruits/satteri-*`
packages. Regenerating the lockfile would not change this; `^0.9.1` forces the
same resolution. Nothing gates satteri:
`tools/test_browser_gate_subset.py` scopes its pin guard to
`@astrojs/markdown-remark` only.

**4. The `@astrojs/markdown-remark` pin gate — checked, clean.** All four values
agree: astro 7.2.4's exact optional peer is `7.2.4`, the `docs-site/package.json`
pin is `7.2.4`, the lockfile resolves `7.2.4`, and Starlight's `^7.2.0` peer
still accepts it.

**5. mermaid 11.17.0 changes the one published diagram.** See *User-visible
change* below.

**6. Net-new packages and their provenance.** `fastdom` 1.0.12, `strictdom`
1.0.1, and `undici` 8.10.0 are net-new to `docs-site/`. **Only `fastdom` reaches
the published client bundle** — measured against the emitted tree, not inferred
from the lockfile:

- `fastdom` is present in `build/docs/_astro/chunk-GMAD6QVW.BD_evIy5.js`,
  imported by `Footer.astro`'s script chunk and every mermaid diagram chunk.
- `strictdom` is **absent** from every emitted JS file: none of its distinctive
  literals (`Can only get `, `Can only set `, `getPropertyCSSValue`,
  `create prototype properties`) appears anywhere in `build/`. It is reachable
  only through `fastdom/fastdom-strict.js`, which is not fastdom's `main`
  (`"main": "fastdom.js"`) and which mermaid does not import. It is a
  build-tree-only transitive.

An earlier revision of this PR body claimed both packages reach visitors'
browsers. That took a lockfile edge for bundle reach without measuring the
bundle; corrected here.

| Package | Published | Maintainers | Provenance attestation |
| --- | --- | --- | --- |
| `strictdom@1.0.1` | 2016-08-05 (3 versions total) | 1 | **no** |
| `fastdom@1.0.12` | 2024-02-20 | 1 | **no** |
| `undici@8.10.0` | 2026-08-03 | 3 | yes |
| `mermaid@11.17.0` | 2026-08-19 | 5 | yes |
| `satteri@0.10.5` | 2026-08-19 | 1 | yes |
| `astro@7.2.4` | 2026-08-19 | 2 | yes |

**Accepted risk, with a retirement condition.** **One** package newly reaching
visitors' browsers lacks build-provenance attestation: `fastdom@1.0.12`
(single-maintainer, 2024-02-20; it does carry a registry signature).
`strictdom@1.0.1` — the 2016 release — also lacks attestation but never reaches
the browser, so its exposure is build-time only. This was put to the repository
owner with the measured figures above; the decision was to record it and land the
bump. The correction narrows the accepted risk rather than widening it.
Following the shape `tools/npm-audit-allowlist.toml` requires — a blank
`unblocked_when` is a tool error there, because "an allowlist that decays into an
undocumented mute list is worse than no allowlist":

- `fastdom@1.0.12` — `sha512-LB+xjSTEbjHE1cWsxu+tN2Xqr1kpi+V9aADI7sVM5ZMaXyYGPHULQMzpJMYqOTULK/73pUkWVzzObFRBkPr+hg==`
- `strictdom@1.0.1` — `sha512-cEmp9QeXXRmjj/rVp9oyiqcvyocWab/HaoN4+bwFeZ7QzykJD6L3yD4v12K1x0tHpqRqVpJevN3gW7kyM39Bqg==`
- **Retired when** `fastdom` publishes with provenance, or mermaid drops the
  dependency, or the `sast-javascript-coverage` lens lands and covers the bundle.
  (`strictdom` is build-tree-only; it is recorded for completeness, not as
  browser-reaching exposure.)

`undici` adds no new outbound-HTTP capability class, and the mechanism is
measured rather than inferred: `unifont` 0.7.4's `{css-tree, ofetch, ohash}`
becomes 0.7.5's `{css-tree, ohash, undici}` — a substitution. In `web/`, `undici`
stays at 8.9.0 but flips from `dev: true` to a production edge via that same
change. Both sites self-host fonts via `@fontsource` with no remote provider
configured, so no new outbound call is exercised.

**Release-age exposure, stated rather than implied.** `mermaid@11.17.0`,
`satteri@0.10.5`, and `astro@7.2.4` were all published 2026-08-19, nine days
before this landing, and they are the packages carrying new code into both
bundles. `.github/dependabot.yml` configures no `cooldown:` and CodeQL is
Python-only. The one real mitigation here is that `tools/audit-npm.py` queries
the live advisory database at merge time — and it did, confirming its endpoint
live via the `lodash@4.17.11` canary. Whether to adopt a `cooldown:` policy is a
separate owner decision, not a gap in this change.

**7. astro's base path-segment stripping fix.** Both sites configure `base`
(`/agent-ready-repo` and `/agent-ready-repo/docs`), which sit on a clean
path-segment boundary, and both are statically built rather than served by
astro's router. Evidence: `tools/check-rendered-site-links.py --build-dir build`
exit 0 over **69,567 emitted links across 283 pages**, clean.

**8. `astro preview` agent-daemonization.** astro 7.2 extends agent-triggered
daemonization to `astro preview`; `web/playwright.config.ts` records that the
browser gate survives only because it sets `ASTRO_PREVIEW_BACKGROUND`, and that no
CI variable appears in `am-i-vibing`'s detector list — so a green gate in a
non-detected shell would prove nothing. Evidence: `make web-browser-gate` exit 0
(182 passed, 0 skipped) **with** an agent-detection probe from `web/` returning
`{"isAgentic":true,"id":"claude-code","name":"Claude Code","type":"agent"}`. The
environment is agent-detected, so the insulation was genuinely exercised and
still holds at 7.2.4.

**9. `vitest` 4.1.11.** Patch bump to `web/`'s runner. Evidence:
`npm test --prefix web` 18/18 files, 129/129 tests, zero skips, run after
rebuilding both sites.

**Ruled out with evidence, not assumed:**

- **Install-script allowlist.** The `hasInstallScript` set is
  `{esbuild@0.28.1, fsevents@2.3.3}` before *and* after in both lockfiles — the
  native-binary packages are prebuilt. No `allowScripts` edit needed, and
  `lint-npm-allow-scripts` confirms it at exit 0.
- **A new native execution surface.** All nine `@bruits/satteri-*` binaries and
  `satteri` were already resident at 0.9.5 on `main` in **both** lockfiles. This
  moves an already-native component's version; it does not introduce native code
  into an all-JS build.
- **Typosquat / registry substitution.** Every entry in both post-bump lockfiles
  resolves to `registry.npmjs.org` with a `sha512` integrity hash. Publisher
  identity was read separately from the registry (`maintainers`), not inferred
  from the lockfile, which records `resolved` and `integrity` only.
- **Lockfile staleness.** `main` has not touched either manifest since the merge
  bases, so the Dependabot resolutions apply unchanged.

## How do you know it works?

| Check | Result |
| --- | --- |
| Emitted HTML, 283 pages, vs same-source pre-bump baseline | **0 regressions** (2 byte-identical, 281 differing only in asset hashes + generator version) |
| Non-HTML: page→asset referent graph | **0 changed pages** |
| pagefind indexed URLs | 232 → 232 |
| sitemaps | no referent-set change |
| `npm test --prefix web` | 18/18 files, 129/129 tests, **zero skips** |
| `tools/lint-npm-allow-scripts.py --root .` | exit 0 |
| `tools/audit-npm.py --root .` | exit 0, advisory endpoint confirmed live |
| `tools/check-rendered-site-links.py --build-dir build` | exit 0 — 69,567 links across 283 pages, clean |
| `make lint-ruff` | exit 0 |
| `make lint-mypy` | exit 0 — 125 source files |
| `make build-check` | completed with `complete — every leg of this target was invoked, SAST/SCA included.` |
| mermaid, real browser at 11.17.0 | all invariants hold, 0 console errors |

**The headline result.** Both markdown renderers moved and the emitted HTML is
structurally identical on every one of 283 pages. That is the question no release
note could answer.

**Scope of that claim.** Every comparison above was measured on **macOS /
darwin-arm64**, using `@bruits/satteri-darwin-arm64`. Both lockfiles carry nine
per-platform native binaries, and the deployed site is built on `ubuntu-latest`
with `satteri-linux-x64-gnu` (`.github/workflows/pages.yml`). So byte-level
equivalence is proven for the darwin binary only. The cross-platform evidence is
weaker but real: the pages `build` job ran green on this head, executing the Linux
build plus the link check, the web vitest suite, and the browser quality gate —
which assert invariants rather than equivalence against a pre-bump baseline.

Two things about how that was measured, because they nearly produced a false
green:

- The comparison had to be **HTML-first with normalisation**. A whole-tree
  content-hash diff drowns 283 meaningful pages in ~495 churning hashed assets
  and a rebuilt pagefind index. And `<meta name="generator" content="Astro
  v7.2.2">` appears on 281 of the 283 pages and moves to `v7.2.4` — unnormalised,
  that is 281 false regressions on its own.
- Bundler chunks carry **two** version-varying tokens. Normalising only the
  trailing one reported 74 assets "gone" and 73 "new"; after normalising both,
  exactly **one** real item remained. The real finding was 1 in 148.

That one item: `docs/_astro/map.<hash>.js`, a 5 KB graphlib/dagre internal chunk,
disappears.

**The mechanical classifier fails on it.** `classify-nonhtml.py` exits **1** with
`"verdict": "REGRESSION"` and `"findings": ["1 hashed asset identity(ies)
disappeared"]`. AC7 ends "Every regression is fixed, not accepted", so that
verdict was overridden by judgement, and this PR says so rather than presenting
the run as clean.

The basis for the override, stated so it can be checked:

- **Zero references to any `map.*.js` remain** anywhere in the emitted JS. This
  is the load-bearing evidence.
- Its four baseline referrers (`dagre.*`, `mermaid.core.*`, two `chunk-*`) all
  show content changes, consistent with Rollup inlining the module into its
  consumers.
- The browser observation independently confirms the lazy-import graph still
  resolves at runtime.

**What the referent graph could *not* tell us.** `classify-nonhtml.py` builds the
page→asset graph from `*.html` only, and `map.<hash>.js` was referenced
exclusively from other JS chunks. So "referent graph: 0 changed pages" is silent
about chunk-to-chunk edges by construction and does **not** corroborate this
disposition — the emitted-JS reference sweep does. Cited here with its scope
rather than as blanket reassurance.

The baseline also had to be **re-captured mid-run**: `main` advanced 35 commits
during review (INI-009 merged), adding two published pages, so the original
baseline was no longer same-source. It was rebuilt at current `main` with the
four manifests reverted to pre-bump — 283 pages on both sides.

## User-visible change: the three-loops diagram is laid out differently

mermaid 11.17.0 ships `#8005`, reverting `#7672`'s subgraph-direction behaviour,
and `getting-started/three-loops` is exactly the shape that revert targets: three
`subgraph` blocks each with `direction TB`, joined by three inter-subgraph edges.

Observed in chromium at 1280x900, `waitUntil: 'load'` plus a 2500 ms settle
(mermaid is a lazy dynamic import and is not ready at load):

| Measure | 11.16.1 | 11.17.0 |
| --- | --- | --- |
| nodes / edges / clusters | 10 / 10 / 3 | 10 / 10 / 3 |
| svg viewBox | `0 0 637.90625 993` | `0 0 379.52734375 1204` |
| DOM node order | RE, PE, CO — not source order | **PE, CO, RE — source order** |
| console warnings + errors | 0 | 0 |

All three invariants hold: three labelled subgraph boxes; each holding its nodes
in TB order; and `G3` → `CO1`, `G4` → `RE1`, dashed `findings` → `CO1`. The
diagram now stacks vertically in source order rather than two-up in scrambled
order, and the pre-existing quirk where `Raw idea` overlapped the top border of
its own subgraph is **gone** — that quirk was a plausible symptom of the very
behaviour this release reverts.

This is a real change to a published page, and an improvement. Worth knowing
about rather than discovering.

A green build proves nothing here: the build emits an empty
`.mermaid-diagram[data-mermaid]` placeholder and
`docs-site/src/components/Footer.astro` catches a render rejection and only
`console.warn`s it, so total failure is a blank gap with no page error. Only a
browser can see it.

- **Splitting the two PRs.** The coupling above makes one PR the smaller
  *reviewable* unit even though it is the larger diff.
- **Regenerating the lockfiles.** That would substitute today's registry state
  for the resolution CI already went green on, and break the provenance check.
- **`mermaid` 11.17.2**, which exists. Outside #1113's resolution; taking it
  would break the byte-provenance guarantee.
- **A JS SAST pass over the shipped bundle**, a **satteri pin guard**, and an
  **emitted-markdown regression test**. Each would add or tighten a gate, which
  this change's scope bars. Reported below instead.
- **Upgrading Starlight** to collapse the dual-satteri resolution — in neither
  PR, and it would trigger `docs-site/AGENTS.md`'s own post-Starlight
  re-verification duty.

## Bundled fix (declared)

`docs/architecture/reference.md` read "`web/` uses Astro 7.1.0; `docs-site/`
uses Astro 7.1.0 with Starlight 0.41.4". Actual values were astro **7.2.2** and
Starlight **0.41.7** — the line was already two minors stale *before* this
change, and this change moves astro again. The document declares itself the
normative reference architecture, and root `AGENTS.md` forbids leaving
documented guidance in conflict with code.

Admitted as a Tier 3 bundled fix: one line, same concern (which astro version the
sites use), both values mechanically readable from the manifests. The correction
preserves the document's existing exact-version form rather than rewriting it to
a range, which would be a design call the bundled-fixes tiers fail closed on.
**The staleness is pre-existing and not caused by this bump.**

## Reportable gaps (not fixed here)

1. **No JS SAST/SCA over emitted bundles.** CodeQL is Python-only and no
   JS ruleset exists, so nothing inspects the shipped client bundle — the surface
   ADR-0083 explicitly recorded as "vendored mermaid's entire transitive tree
   into shipped output". Tracked as `sast-javascript-coverage`.
2. **No satteri pin guard.** `tools/test_browser_gate_subset.py` guards the
   `@astrojs/markdown-remark` pin but has no satteri equivalent, so the
   dual-minor resolution in `docs-site/` is ungated.
3. **No provenance/attestation verification anywhere.** `npm audit signatures`
   appears nowhere in the repository; lockfile integrity hashes detect
   post-resolution tampering only, not a malicious upstream publish.
4. **No emitted-markdown regression test.** This PR establishes by measurement
   that a parser swap changed nothing; nothing would catch the next one.
5. **No Dependabot `cooldown:`.** Nothing imposes a minimum release age.

### DIFF
--- a/docs-site/package-lock.json
+++ b/docs-site/package-lock.json
@@ -8,13 +8,13 @@
       "name": "agent-ready-repo-docs",
       "version": "0.0.1",
       "dependencies": {
-        "@astrojs/markdown-remark": "7.2.2",
+        "@astrojs/markdown-remark": "7.2.4",
         "@astrojs/starlight": "0.41.7",
         "@fontsource-variable/inter": "5.3.0",
         "@fontsource-variable/source-serif-4": "5.3.0",
         "@fontsource/jetbrains-mono": "5.3.0",
-        "astro": "7.2.2",
-        "mermaid": "11.16.1",
+        "astro": "7.2.4",
+        "mermaid": "11.17.0",
         "unist-util-visit": "5.1.0"
       },
       "engines": {
@@ -223,9 +223,9 @@
       }
     },
     "node_modules/@astrojs/internal-helpers": {
-      "version": "0.10.2",
-      "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.10.2.tgz",
-      "integrity": "sha512-yt7fMgPYqSM4Tmr+taTW6Per+hjJ8Pk6lA1PAcDyqzOt8HzJ6Kje5WzCxA2Sd+9wsUW7uhkLeoTMK0cXPwH9rQ==",
+      "version": "0.10.4",
+      "resolved": "https://registry.npmjs.org/@astrojs/internal-helpers/-/internal-helpers-0.10.4.tgz",
+      "integrity": "sha512-nozZSy/mKYLqe4YrqbKtdOszedAfXYCtw3wZ0d+CAjz4GqQ4L9rl1ltIL5BlgwmYVinJg/RZ0MgGuWOdlyRZlA==",
       "license": "MIT",
       "dependencies": {
         "@types/hast": "^3.0.4",
@@ -239,12 +239,12 @@
       }
     },
     "node_modules/@astrojs/markdown-remark": {
-      "version": "7.2.2",
-      "resolved": "https://registry.npmjs.org/@astrojs/markdown-remark/-/markdown-remark-7.2.2.tgz",
-      "integrity": "sha512-FGfmK84zSNcrsBd0dl1gXE9JvZYElp8EXQa2jpHVAxG4deGKAp43wspxFupjADJX7MSsMRHwYCnfT6EyVmgeFQ==",
+      "version": "7.2.4",
+      "resolved": "https://registry.npmjs.org/@astrojs/markdown-remark/-/markdown-remark-7.2.4.tgz",
+      "integrity": "sha512-MvspGMynWKAjTe4/lTUdmBPHIFKNVLTCF6UlyWGogTGzNrTvjD+D4n48k7h8swxsEPKHK2TwxkZO7uoaCv1Pow==",
       "license": "MIT",
       "dependencies": {
-        "@astrojs/internal-helpers": "0.10.2",
+        "@astrojs/internal-helpers": "0.10.4",
         "@astrojs/prism": "4.0.2",
         "github-slugger": "^2.0.0",
         "hast-util-from-html": "^2.0.3",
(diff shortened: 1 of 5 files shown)


eugenelim:
## Verification evidence

**CI on this PR — all five required checks green in a clean environment**

| Check | Result |
| --- | --- |
| `make build-check` | SUCCESS (also windows) |
| `gate-main` | SUCCESS |
| `gate-sast` | SUCCESS — the npm SCA gate, on the combined artifact |
| `gate-export-boundary` | SUCCESS |
| `gate-credbroker` | SUCCESS |

23/24 checks SUCCESS, 0 failing, `mergeStateStatus: CLEAN`. Per
`.github/workflows/build-check.yml`, those five together are CI's equivalent of
`make ci`.

**Local gates**

| Gate | Result |
| --- | --- |
| `make lint-ruff` | exit 0 |
| `make lint-mypy` | exit 0 — 125 source files |
| `make build-check` | `complete — every leg of this target was invoked, SAST/SCA included.` |
| `npm test --prefix web` (after rebuilding both sites) | 18/18 files, 129/129 tests, **zero skips** |
| `tools/lint-npm-allow-scripts.py --root .` | exit 0 — both lockfiles match `allowScripts` |
| `tools/audit-npm.py --root .` | exit 0 — advisory endpoint confirmed live via its `lodash@4.17.11` canary |
| `tools/check-rendered-site-links.py --build-dir build` | exit 0 — 69,567 links / 283 pages, clean |
| `make web-browser-gate` | exit 0 — 182 passed, 0 skipped |

A full local `make ci` was interrupted twice mid-pytest by machine load
(`vm.loadavg` peaked at 282, several peer worktrees active) — not by anything in
this change. Its legs were run individually instead, and the five required checks
above are the authoritative equivalent.

**`astro preview` daemonization insulation — exercised, not assumed**

astro 7.2 extends agent-triggered daemonization to `astro preview`, and
`web/playwright.config.ts` notes that no CI variable appears in `am-i-vibing`'s
detector list — so a green gate in a non-detected shell would prove nothing. The
probe from `web/` returns:

```json
{"isAgentic":true,"id":"claude-code","name":"Claude Code","type":"agent"}
```

The environment **is** agent-detected, so the `ASTRO_PREVIEW_BACKGROUND`
insulation was genuinely exercised and still holds at astro 7.2.4.

**Emitted-output comparison**

Against a **same-source** pre-bump baseline. The first baseline had to be
discarded: `main` advanced 35 commits during review (INI-009 merged), adding two
published pages, so it was no longer same-source and every difference would have
been ambiguous. Rebuilt at current `main` with the four manifests reverted to
pre-bump — 283 pages both sides.

| Class | Result |
| --- | --- |
| HTML, 283 pages | 2 byte-identical, 281 asset-hash/generator-only, **0 regressions** |
| normalisation gap check | 0 — normalisation collapsed the difference rather than hiding a second one |
| page→asset referent graph | 0 changed pages |
| pagefind indexed URLs | 232 → 232 |
| sitemaps | no referent-set change |
| residual + pagefind runtime | 0 findings |
| hashed-asset identities gone / new | 1 / 0 |

The one item: `docs/_astro/map.<hash>.js`, a 5 KB graphlib/dagre internal chunk,
disappears — and **zero references to any `map.*.js` remain** anywhere in the
emitted JS. Rollup consolidated it into its consumers (`dagre`, `mermaid.core`
and two `chunk-*` files all show content changes). No dangling import.

**mermaid, observed in a real browser**

chromium, 1280x900, `waitUntil: 'load'` + 2500 ms settle.

| Measure | 11.16.1 | 11.17.0 |
| --- | --- | --- |
| nodes / edges / clusters | 10 / 10 / 3 | 10 / 10 / 3 |
| svg viewBox | `0 0 637.90625 993` | `0 0 379.52734375 1204` |
| DOM node order | RE, PE, CO | **PE, CO, RE — source order** |
| console warnings + errors | 0 | 0 |

All three invariants hold. The diagram now stacks vertically in source order, and
the pre-existing quirk where `Raw idea` overlapped its own subgraph border is
gone. See the PR body for why this is user-visible and an improvement.

PR #1280, opened 2026-09-13
feat(governance): give every ADR and RFC ordinal a single owner

eugenelim:
## What does this change?

ADR and RFC ordinal uniqueness gains an owner.

The bundled `next-ordinal.py` — shipped to adopters through both `new-adr` and
`new-rfc` — gains a `--check <dir>` mode that reports every ordinal held by more
than one record, and answers `next` from the union of the working tree and the
remote default branch. This repository runs that check fail-closed over
`docs/adr` and `docs/rfc` inside the `build-check` chain, against the projected
copy an adopter would also run.

Four genuine collisions are repaired. The later-landed member of each pair moves:

| Ordinal | Keeps it (landed first) | Moves to |
| --- | --- | --- |
| ADR-0055 | wave-1 docs restructure (07-26 08:39) | Starlight → **ADR-0109** |
| ADR-0106 | direct-skill identity (09-08 20:22) | cooled-child scope → **ADR-0110** |
| RFC-0047 | default-source on discovery verbs (06-25 08:43) | adopter grounding → **RFC-0100** |
| RFC-0074 | fidelity ladder (07-27 23:27) | pack config + oplog → **RFC-0101** |

`CONVENTIONS.md` § 2 gains the rule that makes this legitimate: repairing a
collision is an edit to a record's *identity*, never its content, which is what
reconciles it with ADR immutability. § 3 records the `NNNN-<slug>-research.md`
companion form alongside `NNNN-notes/`.

## Why?

The collision is invisible to review by construction. It exists only against the
default branch and never within the branch that creates it, so no review round
can see it — eight adversarial rounds missed exactly this on ADR-0107. The
allocator reports what is free *now*; it never reserved anything.

The starting report was two ADR collisions. Measuring the corpus found four: the
two RFC pairs had never been reported, and `docs/rfc` also carries twenty shared
ordinals that are legitimate companions and must not fire.

The index had already absorbed both ADR collisions, in three different
improvised ways — a suffixed `0055a`/`0055b` form the conventions define
nowhere; a duplicated bare `0106` with a note asserting neither record would be
renumbered; and, in the RFC index, a dropped row, leaving both keepers with no
entry at all. A column-versus-link check could never have found the `0106` pair,
because those two rows agreed with their own link targets and disagreed only
with each other.

## How was it verified?

Every criterion was driven red before it was accepted green.

- **Record predicate — 7/7 mutations killed.** Both fail-open paths, the symlink
  skip, a swallowed classification error, both companion exclusions and the
  digit boundary were each reintroduced independently; each failed the suite.
- **Post-review fixes — 3/3 killed.** Reverting the `lstat` classification, the
  `ls-tree -z` flag, or the narrowed `resolve()` handler each reds the suite.
- **Gate seeded through both routes.** A duplicate in `docs/adr`, then
  independently in `docs/rfc`. One seeded route leaves the other free to be
  absent or miswired and still green.
- **Real corpus.** A roster suite walks every shared-ordinal group in both
  directories and prints the cross-product: `docs/adr` 0 shared, `docs/rfc` 20
  companions, 0 duplicates. It asserts a non-empty floor first, so a walk that
  reached nothing cannot read as a clean tree.
- **Index check.** The column-versus-link predicate finds 2 mismatched rows on
  the pre-repair index and 0 after, across 110 ADR and 97 RFC rows.
- `make build-check` green. 37 tests pass and 1 skips in each pack suite, 12 in
  the roster suite; `lint-spec-status` clean across 445 specs with no warning
  naming a renamed record; `validate_guides` clean across 203 guides; self-host
  reports no drift across all six projections.

Three defects were found *after* the first green gate, which is the part worth
reading:

1. **The union was a no-op in production while its suite was green.** It ran
   `ls-tree` from the target directory with a repository-root-relative pathspec;
   Git resolves a pathspec relative to the current directory, so it searched for
   `docs/adr/` underneath `docs/adr/` and matched nothing, silently. The fixture
   built its records at the repository root — the one shape where that pathspec
   resolves, and one no real caller has.
2. **The fail-closed check was not fail-closed.** `Path.is_symlink()` and
   `Path.is_file()` return False on any `OSError` rather than raising, so the
   `except OSError` around them was unreachable and an entry that vanished
   between listing and classification was dropped silently. The covering test
   had patched `Path.is_symlink` to raise — proving the handler against
   behaviour the real call never exhibits.
3. **A clean gate was invisible in a build log**, making a step that passed
   indistinguishable from one that was never wired.

- **An allowlist for the known collisions.** It would have let the gate ship
  before the repair, but it converts a defect into configuration and the next
  collision lands green.
- **Trimming the RFC pack suite.** It repeats the ADR suite's git-integration
  matrix against bytes a test already proves identical, and dropping it would
  roughly halve the added CI time. The approved plan has both suites carrying
  the same assertions, so removing coverage is an amendment, not a fix.
- **RFC-0075, RFC-0076 and RFC-0078 are absent from the index.** Found while
  restoring the two rows the collisions had displaced. That gap predates these
  collisions and is unrelated.
- **Moving `0021-greenfield-inception-research.md` into `0021-notes/`.**
  Consolidating to one companion form is a separate governance call; the
  predicate accepts both and § 3 now records both.
- **Merge serialisation.** `main` already requires an up-to-date base and lists
  `make build-check` among its required checks, so the stale-green window is
  closed by existing settings rather than by anything here.

## Review verdict

```json review-verdict.v1
{
  "state": "READY",
  "reviewers": [
    {"role": "adversarial-reviewer", "scope": "spec+plan", "rounds": 3, "result": "clean"},
    {"role": "security-reviewer", "scope": "spec-stage", "rounds": 3, "result": "clean"},
    {"role": "adversarial-reviewer", "scope": "diff", "rounds": 3, "result": "clean"},
    {"role": "quality-engineer", "scope": "diff", "rounds": 1, "result": "findings-applied"}
  ],
  "findings": [],
  "residual_risk": [
    "The RFC pack suite duplicates the ADR suite's git-integration matrix against byte-identical sources; deferred as an amendment rather than a fix."
  ]
}
```


eugenelim:
**Expected warnings from the "ADRs are immutable" check.**

That job is non-blocking by design — it warns and always exits 0, leaving the
call to a reviewer. It will fire here, and this is the note it is waiting for.

It warns when a file that was `Accepted` on the base has body lines changed. Two
groups trip it:

- `0055-starlight-replaces-mkdocs-for-reference-docs.md` and
  `0106-cooled-child-scope-...md` are **renamed**, so at their old paths the
  whole body reads as removed.
- `0050-astro-marketing-site-toolchain-and-deploy.md`, `0085-docs-rendering-is-site-local.md`
  and others have **citations rewritten** where they pointed at a moved record.

Neither changes a decision. `docs/CONVENTIONS.md` § 2 now states the rule this
PR is applying: a duplicate ordinal is a defect in a record's *identity*, not its
content, and repairing it is the one sanctioned edit to an Accepted ADR. Each
moved record carries a `Renumbered:` line naming its former ordinal, and its
decision text is byte-for-byte unchanged apart from its own heading.

What is worth a reviewer's attention instead is **citation attribution**. Both
members of each collided pair are cited as a bare `ADR-NNNN`, so every occurrence
was read and attributed rather than substituted — `ADR-0055` means the wave-1
docs restructure in `tools/lint-build.py` and the Starlight decision in
`tools/build-site.py`. The index-column and link-resolution checks catch a dead
path; **neither catches a citation rewritten to the wrong sibling**, so that
remains the one thing a machine did not verify here.

PR #1289, opened 2026-09-13
feat(frontend-engineering): make the rendered page actually get looked at, and make a defect cost something

eugenelim:
## What does this change?

The `frontend-engineering` pack now opens the page it just built, looks at it,
and writes down what it saw — and a surface can no longer complete over a
reader-visible failure it found. The independent `frontend-reviewer` is handed
those captures and can take its own, so the check that was blind to "one element
covering another" no longer is.

## Why?

- Implements: `docs/specs/rendered-page-visual-inspection/spec.md` (51 criteria)
- Follows from: `docs/product/intents/rendered-page-visual-inspection.md`

The intent named five failures. The pack *asked* for visual QA, but nothing
executed it, nothing recorded it, nothing required it, and the reviewer could not
cover it. This closes all five. Scope was widened mid-delivery by owner decision
to cover the last two, which the approved spec had scoped out.

## How do I verify it?

```bash
python3 -m pytest packs/frontend-engineering/tests/skills/frontend-engineering/ -q   # 235
make lint-ruff lint-mypy
python3 -m agentbundle catalogue lint --root . --deep && python3 -m agentbundle catalogue verify --root .
python3 tools/lint-pack-test-boundary.py --root .   # 8/8; the new suite is registered
```

Beyond the suite, the step was **run for real** rather than asserted: 20 captures
from the shipped fixtures through headless Chromium, judged, with the outcome in
`notes/verification-ledger.md`. Both `Blocker`-class defects report
`completed`/`fail` and are not completed inspections; the two `Major` defects
report without blocking. A separate visual-fuzz pass over 24 generated pages
(87 captures) found zero invariant failures.

**Two pack releases:** `frontend-engineering` 0.2.3 and `core` 2.25.23.

- **Pinning `npx playwright`.** A security round asked for it; adjudication
  refuted it. The same unpinned `npx` pattern is pre-existing pack content, and
  pinning would collide with the `Ask first` on runtime dependencies and the
  criterion that dependency surfaces gain nothing.
- **Sandboxing the reviewer's `Bash`.** The containment gap is the platform's:
  every Bash-capable agent here carries generic `Bash` with no sandbox, and the
  two peer reviewers carry no no-write clause at all. `frontend-reviewer` is now
  the strictest of them. Stated, not enforced.
- **Hardening the rate guard further.** It drew opposite complaints in two
  consecutive rounds, which is the signal that a regex over prose cannot decide
  rate-versus-count. Adjudicated SHRINK: the claim was narrowed instead.
- **A `yes|no` domain check on `page-scrollable`**, **required diagnostic context
  on every failure state**, and **narrowing the spec's Testing Strategy wording** —
  each refuted or declined by the owner, with reasons in the ledger.

---

## Residual risk, stated plainly

**The final round of repairs is unreviewed.** Five review rounds ran; the retry
budget reached 5 of 5. The owner was shown that position *before* the last retry
was spent and chose to fix and ship. Nine round-5 findings plus four from a
mechanical sweep were fixed after the last review.

Mitigating it: every fix is pinned by a check driven against a mutated source,
and both sweeps were re-run afterwards.

**The recurring defect class.** Four times in this delivery a rule landed in the
rule layer and the checks but not on the surface an adopter reads — including
once inside the repair written to fix it. It is now tested mechanically rather
than watched for: delete each shipped rule row and each must fail something;
corrupt each value and the same. That sweep takes six minutes and belongs after
any rule-layer change.

## Review verdict

```json review-verdict.v1
{
  "schema": "review-verdict.v1",
  "state": "READY_WITH_RESIDUAL_RISK",
  "rounds": 5,
  "reviewers": {
    "adversarial-reviewer": "clean at round 3; rounds 4-5 sustained and fixed",
    "security-reviewer": "2 rounds, 2 sustained and fixed, 6 refuted",
    "quality-engineer": "3 rounds, 9 sustained and fixed, 3 refuted"
  },
  "residual": [
    "Round-5 repairs ship unreviewed; review retry budget exhausted at 5 of 5, by owner decision",
    "frontend-reviewer's no-write constraint is stated in its contract, not enforced by the platform",
    "The no-published-rate guard's reach is its stated vocabulary at its stated adjacency form; a rate phrased outside either is an accepted blind spot"
  ]
}
```


eugenelim:
---

## Update — round 6 (authorised past the retry cap)

Six further findings, all sustained. One mattered more than the rest.

**A ticked criterion was false.** The reference's own `Result states` table still
graded `completed` as a completed inspection unconditionally. `SKILL.md` and the
guide had been corrected a round earlier; the rule layer they derive from had
not. Fixed by splitting the two questions: the column is now `Execution
complete` carrying machine-readable `yes`/`no`, and the conjunction — a completed
inspection is execution complete **and** a `pass` verdict — is stated in prose
beside it on all three surfaces.

**Four stale adopter surfaces** the earlier sweep never looked at: the pack
reference page (still five lenses), the audit how-to (never ran the inspection),
the tutorial (screenshot-only gate), and journey step 5 (manifest inventory
omitting fields its own gate requires three steps later).

**The seed is now derived, not typed.** Any shipped file naming both `known
exceptions` and `unverified items` — keys predating this delivery — is a surface
the inspection result must reach. That immediately found a fifth nobody had
named: `fe-status`, the skill whose entire job is reading the manifest and
reporting a surface's quality state, which knew nothing of the inspection and
would have answered without the verdict.

**The shipped capture command could not scroll.** `npx playwright screenshot`
is at-rest only, so half the required capture set was unreproducible from the
instructions. Section 5a now scrolls, records the *attained* offset rather than
the requested one, and asks whether the page scrolls at all.

238 tests. Round 6's fixes ship unreviewed, like round 5's, by owner decision.

PR #1305, opened 2026-09-14
docs(intents): bound every value the repo_backlog projection emits

eugenelim:
## What does this change?

Adds one `Draft` repository intent. It records that `workspace-status` projects
two values from `[backlog].open` with no validation — `summary` and `needs` —
and bounds the work to fixing that one emitter. No code changes.

## Why?

The cause is an ordering gap, not a design decision:

| Date | PR | What it did |
|---|---|---|
| 2026-08-10 | #913 | Added `_repo_backlog_entry_dict`, the backlog projection |
| 2026-08-12 | #928 | The RFC-0083 public-safety sweep over every other emitter |

In #928's diff the backlog emitter appears **only as unchanged context** — all
three mentions are context lines, no `+` or `-` on its body. The sweep reached
the emitters its seed contained, and the newest one was not among them.

Measured against `main` @ 003709805:

- 156 open backlog entries project `summary` verbatim. Longest is **785
  characters**, against a documented limit of 1–500 that
  `extract_repo_backlog` never enforces — no type check, no length bound, no
  charset rule.
- 11 entries project `needs`; 9 are raw dictionaries. Five sibling emitters in
  the same module (`workspace_status.py:178,180,188,190,198`) run theirs
  through `_public_needs` first.

This surfaced while fixing #1304, which corrected the *contract* for the
initiative `name`/`milestone` redaction. That redaction is deliberate and
test-pinned; this is the path the same sweep missed.

## How do I verify it?

```bash
# the ordering claim
git show 4f5b978fa -- packs/core/.apm/skills/workspace-status/scripts/workspace_status.py \
  | grep -n "_repo_backlog_entry_dict"      # 3 hits, all context lines / hunk headers

# the measurements
python3 packs/core/.apm/skills/workspace-status/scripts/workspace_status.py status --root . \
  | python3 -c 'import json,sys; ob=json.load(sys.stdin)["repo_backlog"]["open"]; \
print(len(ob), max(len(e.get("summary","")) for e in ob))'   # 156 785
```

Gates on this branch: `check-release-impact` pass (1 file, none
release-impacting), `make lint-ruff lint-mypy` clean, `catalogue verify: ok`.

**Calling `_public_needs` on the backlog path.** This is the obvious repair and
it is wrong. `_public_need` accepts only strings and returns the sentinel for
anything else — verified by direct call, a `{kind, path, type}` dict returns
`"workspace.toml"`. It would redact all 9 dependency records and destroy the
data `SKILL.md` forbids recovering by rereading raw TOML. The intent records
this explicitly so the next person does not try it.

**Setting `Status: Accepted`.** The intent's own riskiest assumption — that
bounding the projection is the control that matters, rather than the
instruction-inert rule the repository already carries — is unsettled. The intent
states it should close rather than proceed if `workspace.toml` is only ever
review-gated content. Accepting first would commit to work the artifact itself
flags as possibly unwarranted.

**Registering a pointer in `workspace.toml`.** Not requested by the intake flow,
and registering a residual is not a disposition.

**Two adjacent gaps in the same projection,** named as out of scope in the
intent's boundary: no pack/profile/adapter/skill inventory in the output, and no
declared workspace version marker read anywhere.

## Review

`shaping-reviewer` in `intent` mode. First dispatch returned
`MALFORMED(riskiest-assumption)` — condition 3 of 6, verified against
`shaping-reviewer.md:29` before acting on it. An `Assumptions` section was added
and a fresh reviewer dispatched, since that field is material and resuming the
same agent would have been warm review. Second dispatch returned no tokens.


eugenelim:
Closing unmerged: this intent's own kill condition fired.

The intent named its riskiest assumption as "bounding the projection is the control that matters", and stated that it should close rather than proceed if `workspace.toml` is only ever review-gated repository content. The owner has now settled the threat model: `workspace.toml` is working material for developers in the same repository, carrying the same threat profile as source code.

Under that model the prompt-injection motivation does not hold — a hostile `summary` is the same class of problem as a hostile line of source, caught by review. The `needs` asymmetry likewise stops being a defect: the five filtered emitters are arguably the over-built ones rather than the unfiltered sixth.

One residual survives, as a correctness matter rather than a security one: the schema guide documents `summary` as 1-500 characters, `extract_repo_backlog` enforces no bound, and this repository projects a 785-character summary. That is doc-versus-code drift and is tracked separately.

A short comment in `_repo_backlog_entry_dict` will record that the unfiltered pass-through is deliberate, so the next reader does not re-file this finding.

PR #1308, opened 2026-09-14
docs(intents): show an initiative's name and milestone at orientation

eugenelim:
## What does this change?

Adds one `Draft` repository intent proposing that `workspace-status` stop
redacting `initiatives[].name` and `initiatives[].milestone`. No code changes.

## Why?

Orientation currently lists bare slugs. On this repository that is four active
initiatives — `ini-002`, `ini-003`, `ini-007`, `ini-009` — each with a name and
a milestone sitting in `workspace.toml` that the reader cannot see. Naming the
initiative is most of what that section is for.

The redaction was added by the pass that bounded every public value in this
projection, when display prose from `workspace.toml` was treated as untrusted
input. #1304 corrected the skill contract to stop promising values the
projection never emits, which removed the visible defect — a session no longer
prints `ini-002 — workspace.toml (milestone: workspace.toml)`. That left the
cost without the benefit.

The owner has since settled the threat model: **`workspace.toml` is working
material for developers in the same repository, with the same threat profile as
source code.** Under that model the redaction guards nothing. The same
reasoning has already been applied to the sibling backlog projection, which
passes its display prose through as written (#1307).

## How do I verify it?

```bash
# what orientation shows today
python3 packs/core/.apm/skills/workspace-status/scripts/workspace_status.py status --root . \
  | python3 -c 'import json,sys; d=json.load(sys.stdin); \
print([(i["slug"], i["name"], i["milestone"]) for i in d["initiatives"]])'
# -> every name and milestone is the literal "workspace.toml"

# what the file actually holds
grep -A2 '^\["ini-002"\]' workspace.toml
```

`check-release-impact` passes (1 file, none release-impacting).

**Setting `Status: Accepted`.** The intent's riskiest assumption is unsettled
and is deliberately the one that could kill it — see below. `Accepted` needs an
explicit owner decision, not a clean review.

**Making the change directly.** It reverses a shipped control pinned by two
tests. That earns a spec and a `security-reviewer` pass, not a quiet revert,
even when the threat it guarded is out of scope.

**Touching the backlog projection's `summary` and `needs`.** Settled and
recorded in code by #1307; named as out of scope in this intent's boundary.

**The documented-limit drift** (`summary` documented 1–500, enforced nowhere,
785 observed). Still open, still not a security matter, not this intent's
subject.

## The risk this intent is carrying

The redaction ships to adopters. Restoring the values is safe *here* because
the people who write this `workspace.toml` are the people who own and review
this repository — a property of this repository, not one the pack can check.

The kill condition is explicit in the artifact: if any supported install route
puts a `workspace.toml` in front of an agent that the local developers did not
author — generated, vendored, or carried in from another repository — then the
restore belongs to that route's trust decision and this intent should narrow
rather than proceed as written. Answering that is the next step, before any
spec.

## Review

`shaping-reviewer` in `intent` mode, clean on first dispatch — no `MALFORMED`
tokens against its six conditions.


eugenelim:
Amended: the riskiest assumption is settled, and it settles toward proceeding rather than narrowing.

The intent previously carried a kill condition asking whether any install route puts a `workspace.toml` in front of an agent that the local developers did not author. The owner has answered, and the answer needs no route-by-route enumeration: `workspace.toml` lives inside the repository it describes, so any developer who can reach it already has access to that repository's source. A route that could supply a hostile `workspace.toml` could supply hostile source in the same breath, which is the threat the repository already accepts and reviews for.

`Assumptions` now records the narrower thing still taken on trust — that no consumer downstream of this projection treats the two fields as though they were already bounded. Nothing in the current contract promises a bound, so a consumer that assumed one would have been reading a value it was told not to render.

`Opportunity` records the settled question directly, so a reader does not have to infer it from `Assumptions`.

Re-reviewed after the amendment (`Assumptions` and `Opportunity` are both material): `shaping-reviewer` intent mode, no `MALFORMED` tokens. Rebased onto current main.

Status stays `Draft`. The `Accepted` transition is the owner's and has not been made.

PR #1349, opened 2026-09-17
fix(product-strategy): repair three false framework citations and errata their source records

eugenelim:
## What does this change?

Three skills cited frameworks that name different parts than the skills declare, so following the citation lands a reader on a source that contradicts the artifact in front of them.

**`define-ux-strategy` — Levy's fourth tenet.** The skill gave the four tenets ending in *killer UX*, the name from the 2015 first edition of *UX Strategy*. The 2021 second edition renames it *frictionless UX* ([O'Reilly ch. 2 ToC](https://www.oreilly.com/library/view/ux-strategy-2nd/9781492052425/ch02.xhtml); the 1st-edition name is on [Levy's own chapter-2 page](https://jaimelevy.com/ux-strategy-the-four-tenets-of-ux-strategy-chapter-2/)). RFC-0063's own prior-art note already gave the second edition correctly — the pack was generated from two other lines of the same RFC.

Step 2(d) also now asks Levy's actual question. It asked whether the vision "includes what killer UX looks like"; the source defines the tenet as weaving the UX through *"all touch points—online and offline—enabling an experience that is frictionless"* ([O'Reilly excerpt](https://www.oreilly.com/content/four-tenets-ux-strategy/)), which is a check a reader can actually fail.

**NN/g's three components** are now cited by NN/g's own names — *vision (or statements of intent)*, *goals and measures*, *plan* — instead of a paraphrase that misdescribed which label belonged to whom.

**`define-content-strategy`'s composite — ten stale surfaces.** An earlier fix repaired the citation line and left the same claim live everywhere else: the pack README, both pack `DESIGN.md` files, `JOURNEY.md` and its generated web projection, `docs/product/journeys/`, three guide pages, and three assertions in the skill's own `evals.json`. Neither published quad names those parts — Halvorson and Rach's 2012 quad is *Substance / Structure / Workflow / Governance*, and [Brain Traffic's 2018 revision](https://www.braintraffic.com/blog/new-thinking-brain-traffics-content-strategy-quad) is *Editorial Strategy / Experience Design / Content Structure / Process Design*. All ten now name it as the pack's own composite.

**The generator gets the fix too.** RFC-0063 and ADR-0053 carry the misattribution in the decision records the pack was built from, so both take an `## Errata` section rather than an edit to their preserved bodies. `packs/product-strategy/DESIGN.md` now states the rule whose absence caused all three: where a skill's parts are its own composite rather than the source's, it says so at the point of citation.

Two ride-alongs, same area and same concern: an italic aside in `route-it-to-work.md` that rendered with three unbalanced emphasis runs (pre-existing at `HEAD`, caused by bold nested inside the italic run), and the regenerated web journey projection.

## Why?

A false citation is worse than no citation. Citations raise reader trust *whether or not they are true* — [Ding et al., AAAI 2025](https://arxiv.org/abs/2501.01303) found the trust increase holds even when the citations are random, and that trust drops significantly when readers actually check. So a wrong one harvests credibility it has not earned, and the bill arrives exactly when a reader does what a citation invites them to do.

## How do I verify it?

```bash
make lint-ruff lint-mypy
python3 -m pytest tools/test_validate_guides.py tools/test_guide_typed_asides.py \
  tools/test_guide_ledger_integrity.py tools/test_lint_guides_no_repo_only_refs.py \
  tools/test_catalogue_navigation.py tools/test_build_site_projection.py -q   # 86 passed
python3 -m pytest packs/agent-skill-engineering/tests \
  packages/agentbundle/tests/unit/test_catalogue_skill_spec_lint.py \
  packages/agentbundle/tests/integration/test_catalogue_verify_fixture_checks.py -q  # 316 passed
python3 tools/lint-web-journey-parity.py    # 20 journeys in parity
python3 tools/lint-agents-md.py             # docs lint passed
python3 .claude/skills/work-loop/scripts/lint-spec-status.py --root .
```

Residual sweep — every remaining hit should be a correct statement, an erratum, or a frozen record body the errata correct:

```bash
grep -rn "Halvorson\|killer UX" packs/ guides/ docs/ web/ | grep -vi painkiller
```

Two adversarial review rounds ran. Round 1 raised 9 findings, round 2 raised 8. One round-2 blocker was self-inflicted: the round-1 repair introduced a *new* false NN/g claim, caught and fixed. One round-2 finding was refuted against source; the rest are applied.

- **`docs/specs/product-strategy-pack/spec.md` (Shipped) and `plan.md` (Done), and `docs/product/changelog.md`.** All carry the superseded names. They are the record of what was built, and rewriting them would erase the history the errata exist to explain.
- **RFC-0063 and ADR-0053 bodies.** Accepted records; corrections append as errata, per the house convention in RFC-0062 and ADR-0020.
- **Trigger phrases** — `"Halvorson content strategy"` in the `define-content-strategy` description and `eval_queries.json`. Users type that; the skill has to keep matching it.
- **A citation-audit detector.** Built and measured (500 grep lines → 22 candidates across 137 skills), then dropped at the owner's direction. Whether a citation is *false* needs the external source, so it could never be a gate — only triage.
- **The remaining 136 skills.** The audit beyond `product-strategy` is not in this change.
- **Moving attribution out of skill bodies into pack READMEs.** Proposed follow-up, being scoped separately.


eugenelim:
Superseded by #1355, which contains this commit plus the rest of the citation audit it opened up. Closing in favour of that.

PR #1361, opened 2026-09-18
build(docs-site): take mermaid 12, stating the layout engine and repinning lodash-es

eugenelim:
Supersedes #1320. Closes out the open Dependabot queue: #1275 and #1319 gated
green and are already merged; this is the adaptation #1320 needed.

## What does this change?

Takes `mermaid` 11.17.2 → 12.0.0 in `docs-site/` with the two things the bump
requires to be safe:

- **`overrides: { "lodash-es": "4.18.1" }`.** mermaid 12 bundles ELK, which
  brings `chevrotain@11.1.2`, which pins `lodash-es` at exactly `4.17.23` —
  carrying GHSA-r5fr-rjxr-66jc (high) and GHSA-f23m-r3pf-42rh (moderate). That
  is why #1320 fails `gate-sast`. `npm audit`'s only offered fix was reverting
  mermaid. The tree already resolved `4.18.1` via `dagre-d3-es`, so the override
  dedupes `chevrotain` onto the version this tree already ran rather than forking
  it to something untested here. The lockfile ends with one hoisted
  `lodash-es@4.18.1` and no nested vulnerable copy.
- **`layout: 'dagre'` stated in `Footer.astro`'s `mermaid.initialize`.** mermaid 12
  made ELK the default layout engine, which re-lays out the one published fence
  (`getting-started/three-loops`).

Plus the guard, the roster registration it needs, and the note in
`docs-site/AGENTS.md`.

`tools/npm-audit-allowlist.toml` is untouched and still `allow = []`.

## Why?

`gate-sast` is a required check, so #1320 cannot merge as-is, and the appearance
change would not have been caught by anything: #1320's own `build` job — pages.yml,
which runs the rendered-output assertions and the 237-case browser gate — passed on
unmodified mermaid 12, because nothing in the repository reads a rendered diagram's
geometry.

The bump is **not** itself a security fix. mermaid 11.17.2 audits clean, so taking
12 introduces the advisories the override then neutralises. It is taken for the
sanitizer floor — `dompurify` `^3.3.3` → `^3.4.12`, resolving 3.4.13, on a
client-side render path that sanitizes every diagram label — at a measured cost of
+4 packages (573→577: `elkjs` and four `@chevrotain/*` in, `fastdom` and `strictdom`
out). Holding at 11.17.2 and closing #1320 was the live alternative and is recorded
as a considered decline, not an omission.

## How do I verify it?

Every figure below is measured, not read off release notes.

The three-arm measurement, rendering the published fence in a real browser with the
options `Footer.astro` passes:

| Arm | effective `layout` | effective `look` | effective `theme` | viewBox |
|---|---|---|---|---|
| mermaid 11.17.2 | `dagre` | `classic` | `neutral` | 379.53x1204 |
| 12.0.0, nothing stated | `elk` | `classic` | `default` | 305.74x1185 |
| 12.0.0, as shipped here | `dagre` | `classic` | `neutral` | 339.81x1300 |

Two things follow, and both are written into the code rather than left implicit:

- `layout` is the **only** appearance default that drifts. The release notes'
  `neo` look and `redux-color` theme are not in effect in 12.0.0 — the bundle
  carries those literals, which is why grepping for them misleads. Neither is
  pinned, and `Footer.astro`'s comment says to measure before adding a pin here.
- Stating `dagre` holds the **engine** the diagram was reviewed under. It does not
  reproduce 11.17.2's rendering; geometry changed across the major regardless
  (379.53x1204 → 339.81x1300). The claim is "same engine", not "same pixels".

Gates, each run unfiltered with its own exit code read:

```
make lint-ruff lint-mypy                    exit 0
tools/audit-npm.py --root .                 exit 0   both projects clean
tools/lint-npm-allow-scripts.py             exit 0
tools/lint-agents-md.py                     exit 0   docs-site/AGENTS.md 80/80
tools/test-pages-workflow.py                exit 0
tools/lint-ci-parity.py                     exit 0
npm run test:plugins --prefix docs-site     exit 0   15/15
```

Sites and suites, in pages.yml's order:

```
python3 tools/build-site.py                 exit 0
npm run build --prefix web                  exit 0   50 pages
npm run build --prefix docs-site            exit 0   263 pages
npm test --prefix web                       exit 0   158/158
tools/check-rendered-site-links.py          exit 0   88,629 links clean
npm run test:e2e:gate --prefix web          exit 0   237/237
```

Manual QA, on built output served under the production base path
`/agent-ready-repo/docs/getting-started/three-loops/`, in both themes: the diagram
renders with 10 nodes, 3 clusters, 10 edges and every label present; it does not
overflow the prose column; there are **zero off-origin requests and zero failed
requests**, so `spec/docs-site-design-refresh` AC9's no-runtime-CDN invariant
survives the major. Screenshots were read, not just captured.

The new guard was mutation-checked in both directions: removing `layout: 'dagre'`
fails 1 of its 2 assertions, weakening `securityLevel: 'strict'` fails 2 of 2, and
restoring both passes. It was also verified from the installed 12.0.0 tree that
`sanitizeText` still routes label text through `DOMPurify.sanitize`, so `strict`
retains its 11.x meaning.

`tools/test-pages-workflow.py` pins the `test:plugins` roster by equality. That pin
earned its keep here: it caught the roster edit being silently dropped during the
rebase onto the two merged Dependabot PRs.

- **Adopting mermaid 12's ELK layout deliberately.** Which engine lays out a
  published docs diagram is a design decision on a reader-facing surface, not a
  dependency bump; it would re-gate through design review. Surfaced for the owner,
  not taken here.
- **An `npm-audit-allowlist.toml` entry.** ADR-0083 D3 admits one only where no fix
  exists. One exists, so none was added.
- **Pinning `look` and `theme` as the release notes advise.** Measurement says both
  are already in effect; pinning them would assert the library's own defaults and
  fail for reasons unrelated to this diagram.
- **Adding `@mermaid-js/layout-elk`.** mermaid 12 bundles ELK and this repository
  never called `registerLayoutLoaders`.
- **A committed visual-regression snapshot of the diagram.** The screenshot specs
  are deliberately outside the browser-gate allowlist because they write PNGs into
  tracked `docs/specs/**` paths, which pages.yml AC11 forbids. The measurement
  harness was run instead and its numbers recorded above, so the geometry claim is
  evidenced without adding a gate that fights an accepted constraint.
- **`docs-site/AGENTS.md` beyond three bullets.** It sits exactly on the 80-line cap
  `tools/lint-agents-md.py` enforces. The new entry is funded by pointing the
  `astro dev` trap at `web/AGENTS.md`, which already carries that mechanism in full,
  and by giving `make site-link-check` one home instead of two. An earlier draft
  funded it by compressing the print-CSS bullets and lost real facts — notably that
  native Starlight components are exempt — so that approach was reverted and those
  bullets are untouched here.

### Accepted residuals

- **The forced `lodash-es` resolution is unexercised at runtime.** `chevrotain` is
  reached through `@mermaid-js/parser` for non-flowchart diagram types; the only
  published fence is a flowchart, and stating `dagre` keeps `elkjs` off the executed
  path too. The first published fence of a langium-parsed type re-opens this check.
  Low risk: 4.18.1 is one minor above the pin and is what this tree already ran.
- **The new guard asserts the stated configuration, not the rendered geometry.** It
  cannot catch mermaid renaming or retiring `layout`. Named on the test, in
  `docs-site/AGENTS.md`, and here.


eugenelim:
Closing unmerged: the upgrade this adapts to is not worth taking, so the
adaptation should not land either.

Review measured away the only benefit recorded for the bump. The claim was that
mermaid 12 raises the sanitizer floor (`dompurify` `^3.3.3` → `^3.4.12`). The
*declared* range does rise, but the **resolved** sanitizer does not: this
project's lockfile has `dompurify@3.4.13` both before and after, because `^3.3.3`
already admitted 3.4.13. The security benefit is zero.

What remained was all cost:

| | mermaid 11.17.2 | mermaid 12.0.0 |
|---|---|---|
| advisories | none | GHSA-r5fr-rjxr-66jc (high), GHSA-f23m-r3pf-42rh (moderate), in code published to readers |
| resolved `dompurify` | 3.4.13 | 3.4.13 — unchanged |
| packages | 573 | 577 |
| published diagram | unchanged | re-laid out by ELK unless `layout` is stated |
| features this repo uses | — | none; one `flowchart TB` fence, rendered dagre + classic, which is 11's default |

The remediation was also weaker than first recorded. `chevrotain` is **mermaid
12's own direct dependency** (`chevrotain: "~11.1.2"`, sibling of `elkjs`, which
declares no dependencies at all) — not something ELK drags in — and it pins
`lodash-es` at exactly `4.17.23`. So the override forces a transitive resolution
across a hard parent pin, and nothing in this repository exercises it: `chevrotain`
is reached through `@mermaid-js/parser` for non-flowchart diagram types only.

There is no clean upstream fix to wait a week for: `12.0.0` is the only `12.x`
published, and `chevrotain` ≥12 (which drops `lodash-es` entirely) is outside
mermaid's `~11.1.2` range.

Taking mermaid 12 would also have falsified published adopter guidance —
`guides/architect/explanation/architect-diagram-skill-design.md` states that
Mermaid's default layout engine is dagre and that ELK requires
`@mermaid-js/layout-elk`. Both are false for mermaid ≥12. That is now stale
regardless of what this repository pins, and is filed separately rather than
fixed here.

Superseded by #1320's closure. The two patch bumps in the same queue, #1275 and
#1319, gated green and are merged.

PR #1387, opened 2026-09-21
feat(core): close the loop on five assertions, and move capture off the path

eugenelim:
## What does this change?

`work-loop`'s finish checklist reads as five assertions instead of seventeen
checkboxes: acceptance criteria satisfied, verification passed, review passed
with residuals surfaced, artifacts consistent, repository state valid. The
sixteen retained obligations are grouped underneath with their text unchanged,
so nothing that blocked a release before stops blocking it.

Knowledge capture leaves the critical path. The item requiring every DECIDE
scratch note to be routed is removed; `## Capture` keeps its routing rules and
runs after the loop closes.

## Why?

Follows from RFC-0077 § Errata, added here. D4 named work-loop closeout among
the semantic gates where capture runs, which shipped as a blocking checklist
item. A capture obligation enforced at the completion boundary turns "is this
lesson worth keeping?" into a blocking judgment, and the cheapest way past a
blocking judgment is a low-value note.

## Review focus

Capture is off the critical path but not unbounded: two of the first bullet's
defect destinations — a ride-along dispatched now, the next reviewed unit in
this session — exist only while the session is open. The new prose says to run
capture before the session ends rather than merely after the loop. Is that the
right bound, or should those two destinations move to DECIDE instead?

## How do I verify it?

- `pytest packs/core/tests/skills/work-loop/ -q` — 1232 passed, 5 skipped, 70 subtests, 741.95s
- `pytest packs/core/tests/pack/ -q` — 250 passed, 24.0s
- `pytest tests/roster/test_capture_rename_guide.py tests/roster/test_wave4_durable_outputs_and_release.py -q` — 10 passed, 2.3s
- `make lint-ruff lint-mypy` — clean, 148 source files
- `python3 tools/check-core-release.py --base origin/main --root .` — release is consistent
- `make build-self` — leaves the tree clean, so the three SKILL.md copies agree

- **A new RFC.** The owner chose an Errata entry on RFC-0077 over a superseding
  RFC; the entry records the mechanism, not just the conclusion.
- **The `## Capture` body.** C4 and C5 are byte-compared against
  `docs/specs/ride-along-admission-test/spec.md` § The shipped clauses, so the
  section is reframed around them, never reworded.
- **The pull-request item stays a checkbox**, nested under repository state and
  placed last: `test_pull_request_authoring._finish_item` locates it by the
  literal `- [ ] **Pull request opened` and bounds at the next unindented
  `- [ ]`. Six checkboxes render; five assertions read.
- **The other sixteen obligations.** Regrouped, not cut — a deeper trim is a
  separate decision.


eugenelim:
Answering the review question, from the branch that wrote the four-row table your `## Capture` section now carries (`eugenelim/follow-ons-mgmt`, the work-item capture spec).

**The two session-bound destinations should move to DECIDE.** "Before the session ends" is not a bound a reader can act on, for the same reason this PR removes the checklist item: it names a moment with no gate at it.

The four rows are not alike, and that is what makes the split clean:

| Row | When it must happen |
| --- | --- |
| Generalisable practice → the existing seam | any time; the seam is additive |
| Specific, real, **blocked** → captured | any time; a record is the point |
| Specific, real, **ready now** → dispatched in-session | only while the session is open |
| Specific, **failing the razor** → refused non-silently | any time |

Rows one, two and four are durable — the artifact survives the session. Rows three and four's *dispatch* is not: a ride-along grouped with related fixes, and the next independently reviewed unit, both need an open session with the diff still in hand. Deferring those past the loop does not delay them, it deletes them.

DECIDE is where the session still has both. It is also where the ride-along decision is already made — the `blocker-applied` return edge that row three rides is a DECIDE-stage edge, so routing there rejoins a judgement that is already being taken rather than adding one.

That leaves `## Capture`, post-loop, holding only the rows whose destination is an artifact: the seam, a captured record, a non-silent refusal. Each of those is cheap precisely because nothing is waiting on it, which is the property this PR is buying.

**One caution from our side.** Our spec fixes the declined set as rows two to four and has a close-time ordering that depends on all of them being enumerated together — the cap fires before the first dispatch, and the ordinal that matches a correction to its item comes from that single enumeration. If row three leaves for DECIDE, that enumeration splits across two moments and the ordinal stops being position-stable. Workable, but it is a real interaction rather than a clean lift, and we would rather adjust our § D3 than have you work around it. Happy to take that on this side once you have decided.

**Unrelated, but you will hit it:** this PR bumps `core` to 2.26.24 and our branch is already there against main's 2.26.23. Whichever lands second needs 2.26.25.

eugenelim:
Review question answered — by `eugenelim/follow-ons-mgmt`, and the answer is that my bound is wrong. Recording the outcome so this thread does not read as open.

**Accepted.** "Before the session ends" names a moment with no gate at it, which is the same defect this PR removes from the checklist. The two session-bound destinations belong in DECIDE, where the session still holds both the diff and the `blocker-applied` edge they ride.

**One correction of mine, withdrawn.** I objected that "refused, non-silently" was their spec's vocabulary rather than ours, and that row four was miscategorised. Both objections were against main's C4. Their branch has already amended C4 in all three of its homes — the `## Capture` bullet, the `C4` constant in `packs/core/tests/pack/test_ride_along_admission_test.py`, and the blockquote in the Shipped `docs/specs/ride-along-admission-test/spec.md`. I verified that against their branch head rather than taking it on authority. "Refused, non-silently" is the amended text of our own clause; main's "discarded" is the stale side. I also said the split was *blocked* by that amendment cost. It was not blocked — the cost was already paid. It was out of scope for this PR, which is based on main and cannot reach unmerged text.

**One correction of mine that stands, and they took it.** Their row three conflates two destinations — the ride-along dispatch and the next independently reviewed unit. The split moves both.

**Ownership.** `follow-ons-mgmt` takes the C4 split, after their current loop. It carries the row-three split, the actionable bound replacing mine, the move to DECIDE, the Shipped-spec amendment with its re-pins, and their §D3 ordinal adjustment — which is the reason it is cheaper there: the ordinals disagree unless it moves in one change.

**What this PR still ships, unchanged.** The five-assertion closeout, and knowledge capture off the critical path. The soft bound is a known weakness with a named owner rather than a silent one.

Whoever rebases second resolves `discarded` versus `refused, non-silently` in favour of theirs. This PR does not touch the C4 bullet — the only `## Capture` edit here is the intro paragraph above it.

